PR #166, opened 2026-07-25
feat: add a transaction scope guard on the Transactional wrapper

### LINKED ISSUE #161: Add a transaction scope guard to the Transactional publisher wrapper
### Problem

`TransactionalPublisher` exposes `begin_transaction` / `commit` / `abort` on `&self`, so the call order is unconstrained at the type level. All current broker implementations resolve misuse at runtime, and mostly silently: `commit` or `abort` with no open transaction returns `Ok(())`, a second `begin_transaction` either errors or silently joins the outer transaction, and a `publish` issued before `begin_transaction` on a transactional publisher goes out immediately instead of being buffered. A caller that forgets `begin_transaction` still sees a successful `commit`.

The runtime already has the `Transactional` wrapper (`src/runtime/publish.rs`), but the begin -> publish -> commit sequence on it remains imperative and order-unchecked.

### Proposed solution

```rust
let tx = TypedPublisher::new(publisher).transactional();

let mut scope = tx.begin().await?;   // runs begin_transaction; the scope is the only handle
scope.publish(&order).await?;
scope.publish(&receipt).await?;
scope.commit().await?;               // consumes the scope

// commit without begin, double begin, and publish after commit
// no longer exist as expressible programs;
// dropping an uncommitted scope aborts on a best-effort basis
```

The scope borrows the wrapper, so the guarantee is per-call-site; the underlying trait stays `&self` as the low-level broker SPI.

### Alternatives considered

- Consuming-self typestate on the `TransactionalPublisher` trait itself (`begin(self) -> InTransaction`). Rejected: broker publishers are `Clone` over shared state, are handed to handlers through `State` DI, and at least one broker design opens a transaction in one task, commits in another, and lets concurrent publishers join it. The trait cannot own the transaction.
- Documentation only: keeps every misuse silent.

### Additional context

Additive and zero-cost (monomorphic, no heap), can ship in a 0.5.x patch. Pairs with a separate breaking proposal that tightens the `TransactionalPublisher` runtime contract itself. Drop-abort must be best effort (no blocking in destructors, per the crate conventions).

powersemmi:
# Description

Adds a manual transaction API on the `Transactional` publisher wrapper: `begin()` opens a `TransactionScope` that owns the broker transaction. Publishes go through the scope, and `commit()` / `abort()` consume it, so a commit without a begin, a double commit, or a publish after settling are compile errors instead of the silent runtime no-ops the broker implementations currently exhibit.

The scope encodes values with the wrapper's codec and sends them directly; reply transforms and the app-wide publish middleware remain dispatch-path features (documented). A failed commit aborts on a best-effort basis so the handle is not left wedged; dropping an unsettled scope logs a warning, since destructors cannot run async work (a deviation from the issue sketch, which suggested best-effort abort on drop). Additive only: the `TransactionalPublisher` SPI is unchanged.

Fixes #161

## Type of change

## Checklist

      `examples/` where applicable

### DIFF
--- a/examples/publishing.rs
+++ b/examples/publishing.rs
@@ -6,7 +6,9 @@
 //! cargo run --example publishing --features macros,memory,json -- run
 //! ```
 
+use std::error::Error;
 use std::future::Future;
+use std::io;
 use std::pin::Pin;
 
 use ruststream::codec::{Codec, JsonCodec};
@@ -15,7 +17,7 @@ use ruststream::runtime::{
     App, AppInfo, HandlerResult, Outgoing, PublishLayer, PublishNext, PublishTransform, RustStream,
     TypedPublisher,
 };
-use ruststream::{OutgoingMessage, Publisher, subscriber};
+use ruststream::{OutgoingMessage, Publisher, TransactionalPublisher, subscriber};
 use serde::{Deserialize, Serialize};
 
 // A publisher shared with handlers as a typed field of the application state.
@@ -95,9 +97,7 @@ impl PublishLayer for AuditPublish {
         &'a self,
         out: &'a mut Outgoing<'a>,
         next: PublishNext<'a, N>,
-    ) -> Pin<
-        Box<dyn Future<Output = Result<(), Box<dyn std::error::Error + Send + Sync>>> + Send + 'a>,
-    > {
+    ) -> Pin<Box<dyn Future<Output = Result<(), Box<dyn Error + Send + Sync>>> + Send + 'a>> {
         Box::pin(async move {
             println!("publishing to {}", out.name());
             next.run(out).await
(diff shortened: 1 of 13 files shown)


src/runtime/publish.rs
```diff
@@ -804,6 +805,167 @@ async fn abort_quietly<P: TransactionalPublisher>(publisher: &P) {
...
+    /// mid-flight may leave the message in an indeterminate state inside the transaction.
+    pub async fn publish<T: Serialize + Sync>(
+        &mut self,
+        name: &str,
+        value: &T,
+    ) -> Result<(), TransactionPublishError<P::Error>> {
+        let payload = self
+            .codec
+            .encode(value)
+            .map_err(TransactionPublishError::Encode)?;
```
powersemmi, line 911: Is it normal to use "?"

examples/publishing.rs
```diff
@@ -95,9 +97,7 @@ impl PublishLayer for AuditPublish {
         &'a self,
         out: &'a mut Outgoing<'a>,
         next: PublishNext<'a, N>,
-    ) -> Pin<
-        Box<dyn Future<Output = Result<(), Box<dyn std::error::Error + Send + Sync>>> + Send + 'a>,
-    > {
+    ) -> Pin<Box<dyn Future<Output = Result<(), Box<dyn Error + Send + Sync>>> + Send + 'a>> {
```
powersemmi: Better this large return type of function to take out separately

PR #167, opened 2026-07-25
feat: enforce app builder call order with a phase type parameter

### LINKED ISSUE #162: Enforce app builder call order with a phase type parameter
### Problem

`RustStream::on_startup` changes the app state type and rebuilds the builder, discarding starters and hooks registered before it (`src/runtime/app/service.rs`, the `starters: Vec::new()` rebuild), while the handler metadata is carried over. Calling `on_startup` after `with_broker` therefore compiles, starts cleanly, dispatches nothing, and still lists the dropped handlers in the AsyncAPI document. The required order (`on_startup` before any registration, `layer` / `publish_layer` before `with_broker`) lives only in doc comments; violating it produces no error, no warning, and no runtime failure.

### Proposed solution

```rust
// today: compiles, silently drops the earlier registration
RustStream::new()
    .with_broker(broker, |b| { /* handlers */ })
    .on_startup(|| async { Ok(state) });

// proposed: a phase type parameter on the builder;
// on_startup and layer exist only before the first registration,
// so the misordered chain above fails with a method-not-found error
RustStream::new()
    .on_startup(|| async { Ok(state) })
    .layer(tracing_layer)
    .with_broker(broker, |b| { /* handlers */ });
```

The builder already carries three type parameters (`L`, `St`, `PP`); a fourth phase parameter follows the existing pattern and is zero-cost.

### Alternatives considered

- Runtime panic on a misordered call: catches the bug at startup instead of at build, and only when the misordered path actually runs. Better than silence, weaker than the type-level fix.
- Carrying earlier registrations across the state-type change: not possible, the earlier closures captured a different state type.

### Additional context

Breaking (pre-1.0 minor bump, 0.6 candidate); needs maintainer sign-off. Follows the compile-time invariants convention in CLAUDE.md (typestate: no out-of-order calls). Existing precedents in the crate: `RustStream -> RunningApp`, `Router`'s type-carried codec and layer stack.

powersemmi:
# Description

Adds a `Phase` type parameter to the `RustStream` builder. `new()` starts in `Setup`, where `on_startup`, `layer`, and `publish_layer` are available; the first `with_broker*` call moves the builder to `Wired`, where those methods no longer exist. A configuration call that could not apply to already-registered handlers - previously a silent drop of registered starters (with AsyncAPI metadata still listing them), or a layer that silently wrapped nothing - is now a compile error. Two trybuild cases pin the rejection.

`on_startup` additionally panics (documented) if a lifecycle hook was registered before it: hooks close over the state type and cannot be carried across the state change; this is the one ordering hole the two-phase model does not express in types.

`Phase` defaults to `Wired`, so existing signatures that name `RustStream` (which almost always means the built service) keep compiling unchanged; construction sites resolve by inference and are unaffected. `App`, `run`/`start`, and the `TestApp` harness are generic over the phase.

Breaking (pre-1.0 minor bump, 0.6): `with_broker*` now return a phase-changed type, and misordered builder chains no longer compile. No version bump in this PR.

Fixes #162

## Type of change

## Checklist

      `examples/` where applicable

### DIFF
--- a/src/runtime/app/app_trait.rs
+++ b/src/runtime/app/app_trait.rs
@@ -11,7 +11,7 @@ use super::{AppInfo, RunningApp, RustStream, RustStreamError};
 
 mod sealed {
     pub trait Sealed {}
-    impl<L, St, PP> Sealed for super::RustStream<L, St, PP> {}
+    impl<L, St, PP, Phase> Sealed for super::RustStream<L, St, PP, Phase> {}
 }
 
 /// The functional surface of a built [`RustStream`] service: run it, and read the metadata the
@@ -88,7 +88,7 @@ pub trait App: sealed::Sealed + Sized {
 // `St: 'static` mirrors the inherent impl in run.rs: every constructible app already satisfies
 // it, and `start` needs it to bind the state into the shutdown hooks.
 #[allow(clippy::use_self)]
-impl<L: Send, St: Send + Sync + 'static, PP: Send> App for RustStream<L, St, PP> {
+impl<L: Send, St: Send + Sync + 'static, PP: Send, Phase> App for RustStream<L, St, PP, Phase> {
     fn run(self) -> impl Future<Output = Result<(), RustStreamError>> + Send {
         RustStream::run(self)
     }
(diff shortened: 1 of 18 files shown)


src/runtime/app/app_trait.rs
```diff
@@ -11,7 +11,7 @@ use super::{AppInfo, RunningApp, RustStream, RustStreamError};
 
 mod sealed {
     pub trait Sealed {}
-    impl<L, St, PP> Sealed for super::RustStream<L, St, PP> {}
+    impl<L, St, PP, Phase> Sealed for super::RustStream<L, St, PP, Phase> {}
```
powersemmi: Can these types also make human-readable names?

PR #168, opened 2026-07-25
feat: take NonZeroUsize in Buffered max_size and Workers counts

### LINKED ISSUE #164: Use NonZeroUsize for Buffered max_size and Workers counts
### Problem

Zero is representable where it has no meaning, and the two registration paths disagree on how to handle it:

- `Buffered::max_size(usize)` accepts 0 and silently clamps it to 1 at subscribe time (`src/buffered.rs`).
- `Workers::pool(0)` and `Workers::keyed(0)` silently rewrite the count to 1 (`src/runtime/dispatch.rs`).
- The macro path rejects `workers(0)` as a hard compile error, so the same invalid value is a compile error in the attribute form and a silent adjustment in the functional form.

### Proposed solution

```rust
use core::num::NonZeroUsize;

impl<D> Buffered<D> {
    pub fn max_size(self, max_size: NonZeroUsize) -> Self { /* ... */ }
}

impl Workers {
    pub const fn pool(count: NonZeroUsize) -> Self { /* ... */ }
    pub const fn keyed(count: NonZeroUsize) -> Self { /* ... */ }
}
```

The clamps disappear, and the functional path gains the same compile-time rejection the macro path already has.

### Alternatives considered

- Keep clamping: silently changes user intent and the paths stay inconsistent.
- Panic on zero: moves the check to runtime instead of the type, strictly worse than `NonZeroUsize`.

### Additional context

Breaking signature change (pre-1.0 minor bump, 0.6 candidate). Follows the compile-time invariants convention in CLAUDE.md (newtype: parse, don't validate).

powersemmi:
# Description

`Buffered::max_size`, `Workers::pool`, and `Workers::keyed` now take `NonZeroUsize`. A zero bound was previously accepted and silently clamped to one (twice on the buffered path), while the macro form `workers(0)` was a hard compile error - the two registration paths disagreed on the same invalid value. With the parameter type carrying the invariant, the clamps are gone and both paths reject zero at compile time; the macro lowering constructs the `NonZeroUsize` through an unreachable-safe match with no panic path.

Breaking (pre-1.0 minor bump, 0.6): three public signatures change from `usize` to `NonZeroUsize`. No version bump in this PR.

Fixes #164

## Type of change

## Checklist

      `examples/` where applicable

### DIFF
--- a/examples/subscribers.rs
+++ b/examples/subscribers.rs
@@ -5,14 +5,13 @@
 //! cargo run --example subscribers --features macros,memory,json -- run
 //! ```
 
-use std::num::NonZeroUsize;
 use std::time::Duration;
 
 use ruststream::codec::JsonCodec;
 use ruststream::memory::MemoryBroker;
 use ruststream::runtime::{AppInfo, Context, HandlerMetadata, HandlerResult, RustStream, typed};
 use ruststream::subscriber;
-use ruststream::{Buffered, Name};
+use ruststream::{Buffered, Name, nonzero};
 use serde::Deserialize;
 
 #[derive(Debug, Deserialize)]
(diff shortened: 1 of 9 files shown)


examples/subscribers.rs
```diff
@@ -86,7 +87,7 @@ async fn reconcile(orders: &[Order]) -> Vec<HandlerResult> {
 // 20 ms after its first one. The macro recovers the source type from the constructor path, so
 // the generic parameter is spelled out.
 #[subscriber(batch(Buffered::<Name>::new(Name::new("orders"))
-    .max_size(128)
+    .max_size(NonZeroUsize::new(128).unwrap())
```
powersemmi: I don’t really like this type of record, whether it is possible to somehow return the order of the number in the macro, without the explicit transfer of NonZeroUsize

PR #169, opened 2026-07-25
feat: tighten the TransactionalPublisher contract

### LINKED ISSUE #163: Tighten the TransactionalPublisher contract: error on misuse, defined post-failure state
### Problem

The `TransactionalPublisher` contract leaves misuse implementation-defined, and every implementing broker has converged on silent success:

- `commit` or `abort` with no open transaction returns `Ok(())` in all three implementing broker crates, so a caller cannot distinguish "committed N messages" from "there was no transaction at all".
- The state after a failed `commit` is undefined. One implementation stays permanently "open" (every later `begin_transaction` fails until the publisher is replaced); others clear their buffer before the fallible flush, so the documented abort-on-error recovery runs against a transaction that no longer exists and the unsent messages are lost without a signal.
- The runtime itself masks the second case: on a commit error it calls a quiet abort (`src/runtime/publish.rs`), which hits the silent no-op path.
- The trait promises all-or-nothing visibility, while at least one implementation documents best-effort batching ("messages already flushed stay published"). Two different guarantee levels satisfy the same bound.

### Proposed solution

```rust
// contract-level changes, verified by conformance:
// 1. commit/abort with no open transaction returns an error
//    (a dedicated variant, not Ok and not a panic);
// 2. the state after a failed commit is defined: an explicit abort
//    returns the publisher to idle, and a begin after a failed commit
//    either works or returns a dedicated error - it never wedges forever;
// 3. the guarantee level becomes visible in the type system, for example
//    a separate marker trait for atomic (all-or-nothing) implementations,
//    so callers can require the stronger guarantee as a bound.
```

New `conformance` checks drive all three rules against the in-memory broker and real brokers.

### Alternatives considered

- Status quo: misuse stays silent, and transaction bugs surface as data-visibility anomalies in production rather than errors.
- Solving it only in the runtime wrapper (the transaction scope guard proposal): protects users of the high-level API but leaves the SPI contract undefined for broker authors and for direct trait consumers.

### Additional context

Breaking behavior change for the core contract and all broker crates (pre-1.0 minor bump, 0.6 candidate; coordinated across the broker line-up per the versioning policy). Pairs with the additive transaction scope guard proposal (#161).

powersemmi:
# Description

Tightens the `TransactionalPublisher` contract so misuse surfaces as errors instead of silent success:

- `commit` / `abort` with no open transaction must return an error, so `Ok` from `commit` always means "an open transaction committed".
- A second `begin_transaction` while one is open must error and leave the open transaction untouched.
- The state after a failed commit is now defined: the transaction is closed (the implementation discards or aborts broker-side), the handle stays usable, and a later `begin_transaction` must work or error - it must never wedge permanently. The runtime accordingly no longer issues a follow-up abort after a failed commit (in the batch reply path and the transaction scope).

The in-memory broker is the executable reference: its publisher error changes from `Infallible` to a new `MemoryError` (`TransactionBusy` / `NoTransaction`), and `conformance::capabilities::transactions` now drives all three misuse paths, holding broker crates to the same rules.

Expressing the atomic vs best-effort guarantee level in types was considered and deferred: it needs its own design pass and touches broker crates directly, so it stays follow-up work rather than a rider on this contract change.

Stacked on #166 (the transaction scope). Breaking (pre-1.0 minor bump, 0.6) for all broker crates; no version bump in this PR.

Fixes #163

## Type of change

## Checklist

      `examples/` where applicable

### DIFF
--- a/docs/broker-authors/conformance.md
+++ b/docs/broker-authors/conformance.md
@@ -94,7 +94,7 @@ without the capability simply do not call it. Each suite takes the same factory 
 |---|---|---|
 | `capabilities::request_reply` | `RequestReply` | the request reaches a responder with a usable `reply-to` header, the correlated reply resolves the request, an unanswered request fails after its timeout |
 | `capabilities::batches` | `BatchSubscriber` | every published message arrives in publish order, distributed over non-empty batches |
-| `capabilities::transactions` | `TransactionalPublisher` | nothing inside a transaction is visible before `commit`, a commit publishes the buffer in order, an abort discards it |
+| `capabilities::transactions` | `TransactionalPublisher` | nothing inside a transaction is visible before `commit`, a commit publishes the buffer in order, an abort discards it; misuse errors - `commit` / `abort` with no open transaction, a second `begin_transaction` while one is open (which must leave it untouched) |
 
 <!-- inline-rust: worked request-reply capability check against the external ruststream-nats crate; its real gated suite lives in that repo, so it has no compiled home here -->
 ```rust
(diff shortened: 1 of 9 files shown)


powersemmi (commented):
All ok!

PR #236, opened 2026-08-19
feat: unify publishing behind one builder with defaulted positions

### LINKED ISSUE #231: Unify publishing behind one builder with defaulted positions
## Problem

There are two ways to publish one message, wanted for the same job and largely interchangeable.

- `Publisher::publish(OutgoingMessage)` takes bytes and an explicit destination. No encoding, no
  declaration, no checks.
- `TypedSlot::publish_typed(&value)` (with `with_headers(..)` in front when the message declares a
  header contract) takes a value, encodes it with the slot's codec, and reads the destination off
  the marker's `#[publishes(..)]` dictionary. The declared set, the name and the header contract
  are enforced at compile time.

Picking one picks a whole bundle of behaviours at once: where the destination comes from, whether a
codec applies, whether headers are contract-checked, whether the message shows up as a declared
outgoing message in the generated document. A user who wants a typed value at a destination decided
at run time, or bytes with a typed header contract, has to leave the path and rebuild the pieces by
hand. The call-order constraint leaks too: `with_headers(..)` has to come before `publish_typed(..)`
because the terminal cannot be an `IntoFuture` builder at zero cost on the current MSRV.

What a message type says about being sent is also split across two places: the payload side is
declared on the type, while the destination lives in a dictionary on the slot marker. There is no
way to say that a type belongs to a space of names rather than to one name, so a service routing
the same message per tenant or per shard computes the name and drops to the byte path, and the
generated document then says nothing about where that type goes.

## Proposed solution

Everything a type needs in order to be sent is declared on the type, by one derive, with every
parameter in the same `key = value` form. `headers` is optional and names the contract type, which
stays an ordinary serde struct that the derive does not touch.

```rust
#[derive(Outgoing, Serialize)]
#[outgoing(name = "orders.done", headers = OrderMeta)]
struct OrderDone { id: u64 }

#[derive(Outgoing, Serialize)]
#[outgoing(name = "orders.{tenant}.{region}.v1")]
struct OrderPlaced { id: u64 }

// The derive alone, declaring nothing: a type that is simply sent where the caller says.
#[derive(Outgoing, Serialize)]
struct OrderArchived { id: u64 }
```

Publishing is one builder with two entry points, `message(..)` for a value and `raw(..)` for bytes.
Which positions exist is decided by the message type, so a call site cannot be under-specified:

```rust
// Bound to one name: the destination is already resolved.
publisher.message(&done).with_headers(&meta).publish().await?;

// Bound to a space of names: one setter per placeholder, generated from the template.
publisher.message(&order).to().tenant(tenant).region(region).publish().await?;

// Declared with the derive alone: the name is a literal at the call site. This is the typed
// replacement for today's `publish(OutgoingMessage::new(name, payload))`, encoding included.
publisher.message(&archived).to("orders.archived").publish().await?;

// Bytes: same shape, no codec position.
publisher.raw(&bytes).with_headers(&headers).to("orders").publish().await?;
```

Positions and where they come from:

- **Destination**: the type's declaration decides, statically, which form of `to(..)` exists. A
  fixed name has none. A templated name has a `to()` that opens a generated builder with one
  setter per placeholder; `publish()` exists only once every placeholder is bound, and an unbound
  one is carried in the builder's type so the error names the segment, the way an unbound `Out`
  slot already names itself today. A type with no declaration, and the byte entry point, take a
  plain name. In every case the destination is resolved before `publish()` exists.
- **Codec**: most specific wins, along the ladder that already governs decode: the call, then the
  router or broker scope, then the application, then the crate default. The byte entry point has no
  codec position.
- **Headers**: the contract named by `headers = ..`, required when the type names one, absent
  otherwise. The contract type keeps its own fields and its own serde renames; the address never
  imposes anything on it.

Naming: `to(..)` rather than `set_name(..)`, since Rust builders avoid the `set_` prefix and the
the framework's word for a destination is a name; `raw(..)` rather than `raw_message(..)`, matching the
`raw` subscriber clause and the byte input kind; `with_headers` and `with_codec` keep the names they
already have. The terminal is an explicit `publish()`.

`name` is the framework's word for a destination, the one every broker's queue, subject or topic
maps onto, so it is what the attribute uses; the document's own channel object is built from it.
The message component keeps taking its name from the type, as it does today.

Because the name moves onto the message type, the slot marker stops carrying one. Its
`#[publishes(..)]` dictionary reduces to what it is really for, the set of types a slot may
publish, which the `Out<impl Publisher, Marker, Set>` position already expresses.

The compile-time guarantee is total rather than optional. The destination is either fixed by the
declaration or demanded by the builder, and the difference between those cases is a compile error
rather than a runtime check.

## Alternatives considered

- Keep both methods and document when to use which. This is the current state; the two surfaces
  keep growing apart, and the recent per-element header work had to be designed twice.
- Add the missing combinations to the existing methods (a `publish_typed_to`, a raw publish that
  takes a contract). Additive, but the method count grows with the product of the positions, which
  is what the builder exists to avoid.
- Let `to(..)` override a type bound to one name. The document would then state a channel the
  service does not necessarily use, while a space of names is expressible as a space.
- Declare the address as a separate type the user writes and points the message at. It doubles the
  declarations to keep in step, and when that type is shared with the header contract the address
  starts imposing rules on the contract's fields.
- Pass the address as a generated struct literal (`to(OrderPlacedAddress { .. })`) or through a
  call-site macro (`to(address!(tenant = ..))`). Both check completeness, but the first puts a
  generated type name in every call site and the second reports mistakes in terms of the expansion
  rather than the address.
- Take the parameters positionally (`to(OrderPlaced::address(tenant, region))`). Shortest to write,
  and the only form where two adjacent segments of the same type can be swapped unnoticed.
- Write the template at the call site (`to(topic!("orders.{tenant}", tenant = ..))`). The
  declaration would leave the type, and the document could no longer say where the type goes.
- Make the builder terminal an `IntoFuture` so the call reads `publisher.message(&m).await`. Not
  available at zero cost on the current MSRV: the publisher's future is an RPITIT that cannot be
  stored in the builder, and boxing it is not acceptable on this path.

## Additional context

Additive. The builder is added next to what exists; `publish_typed`, the byte publish path and the
name-carrying form of the marker's `#[publishes(..)]` dictionary keep working and are deprecated in
place, so no signature changes and no broker crate is touched. Removing the deprecated surface is a
later minor on its own.

Scope: every publish surface moves to the builder in one go - `Out` slots, the typed publisher,
the transaction scopes and the batch publishing path - so the surfaces do not drift apart again
while half of them are migrated.

`ErasedPublisher::publish_bytes` and `ErasedPublisher::publish_message` are deprecated in the same
change. The pair exists only because one carries headers and the other does not, which is a
position on the builder rather than a second method; the trait is blanket-implemented for every
`Publisher`, so nothing outside this crate implements it and the deprecation reaches no broker. The
one internal user, the broker-agnostic `retry_after` fallback, moves to the builder with everything
else.

The two entry points of the builder, `message(..)` and `raw(..)`, are the paths a service is meant
to use. `Publisher::publish(OutgoingMessage)` stays exactly as it is, but as the broker interface
rather than a user-facing one: it is what a broker crate implements, and the builder is implemented
on top of it, resolving every position down to one call. `OutgoingMessage` keeps its place as the
value that interface carries. The split is worth stating in the documentation too, so the guides
show the builder and the broker-author pages show what a broker has to provide.

In the generated document a type bound to one name is declared on that channel as it is today; a
type
bound to a template is declared on the templated address, with the parameters block filled from the
template's placeholders, so the declaration and the call site cannot drift apart. A type with no
declaration contributes nothing, which is what it says about itself.

Cost: a templated address is rendered per publish, so that path allocates where a fixed name
does not. That is inherent to computing an address at run time, and the fixed path keeps publishing
from a `&'static str`.

powersemmi:
# Description

Publishing was two surfaces for one job: a byte publish that took a destination and no
declaration, and a typed publish that read its destination off a slot marker's dictionary. Which
one a call site used decided a whole bundle of behaviours at once, and a service wanting a typed
value at a destination computed at run time had to leave the path and rebuild the pieces by hand.

A message type now declares everything about being sent through one derive, with every parameter
in the same `key = value` form: `#[derive(Outgoing)]` with `#[outgoing(name = "orders.done",
headers = OrderMeta)]`. `name` is the destination - a fixed one, or a template whose
`{placeholder}` segments become setters - and `headers` names the contract type, which stays an
ordinary serde struct the derive does not touch.

Publishing is one builder with two entry points, `message(..)` for a value and `raw(..)` for
bytes, on every publish surface (`Out` slots, the typed publisher, both transaction kinds, and any
bare publisher). Which positions a call site has to fill is decided by the message type, so a
publish cannot be under-specified: a fixed name has no `to(..)` at all, a template demands its
placeholders (an unbound one rides in the builder's type, so the error names the segment), and a
type declaring nothing takes a plain name. A declared header contract is demanded the same way.

`Publisher::publish(OutgoingMessage)` is untouched, but stated for what it is: the interface a
broker crate implements, with the builder resolving every position down to one call to it. No
broker crate is affected by this change.

Additive: `publish_typed`, the name-carrying `#[publishes(Type = "channel")]` form and the
`ErasedPublisher` byte pair keep working and are deprecated in place, with their regression
coverage kept. The marker's dictionary reduces to the list of types a slot may publish. In the
generated document a fixed name becomes its channel, a templated one is declared on its templated
address with the parameters block filled from its placeholders, and a type that declares no
destination contributes nothing.

Fixes #231

## Type of change

## Checklist

      `examples/` where applicable

### DIFF
--- a/crates/ruststream-macros/src/lib.rs
+++ b/crates/ruststream-macros/src/lib.rs
@@ -7,6 +7,7 @@
 
 mod expand;
 mod from_ref;
+mod outgoing;
 mod parse;
 
 use proc_macro::TokenStream;
@@ -284,6 +285,9 @@ pub fn derive_message(item: TokenStream) -> TokenStream {
         {
         }
 
+        // The set's channels come from the slot dictionary, whose name-carrying form is
+        // deprecated; the type deriving Message is not, so the warning stays at the dictionary.
+        #[allow(deprecated)]
         impl #set_impl_generics ::ruststream::runtime::OutMessages<__RsM> for #name #ty_generics
         #set_where_clause
         {
@@ -322,6 +326,49 @@ pub fn derive_message(item: TokenStream) -> TokenStream {
         }
     }
     .into()
+}
+
+/// Derives everything a message type declares about being sent, from one `key = value`
+/// attribute: its destination
+/// ([`OutgoingDestination`](../ruststream/trait.OutgoingDestination.html)), its optional typed
+/// header contract, and the [`Message`](../ruststream/trait.Message.html) metadata the generated
+/// document reads (the type's name and doc comment).
+///
+/// The destination decides which positions the publish builder demands at the call site:
+///
+/// * `#[outgoing(name = "orders.done")]` fixes it - `publisher.message(&done).publish()` names
+///   nothing.
+/// * `#[outgoing(name = "orders.{tenant}.v1")]` declares a space of names - `to()` opens one
+///   setter per placeholder, and the publish appears once the last one is bound.
+/// * the derive alone declares nothing - the call site names it: `.to("orders.archived")`.
+///
+/// `headers = Meta` declares the typed header contract: `Meta` stays an ordinary serde struct
+/// the derive does not touch, and the publish builder then demands `with_headers(&meta)`.
+///
+/// Do not combine with `#[derive(Message)]` on the same type: this derive already provides the
+/// message metadata, and the two deliberately produce conflicting impls.
+///
+/// ```ignore
+/// /// A finished order.
+/// #[derive(Outgoing, Serialize)]
+/// #[outgoing(name = "orders.done", headers = OrderMeta)]
+/// struct OrderDone { id: u64 }
+///
+/// #[derive(Outgoing, Serialize)]
+/// #[outgoing(name = "orders.{tenant}.{region}.v1")]
+/// struct OrderPlaced { id: u64 }
+/// // out.message(&placed).to().tenant("acme").region("eu").publish().await?;
+///
+/// #[derive(Outgoing, Serialize)]
+/// struct OrderArchived { id: u64 }
+/// // out.message(&archived).to("orders.archived").publish().await?;
+/// ```
+#[proc_macro_derive(Outgoing, attributes(outgoing))]
+pub fn derive_outgoing(item: TokenStream) -> TokenStream {
+    let input = parse_macro_input!(item as DeriveInput);
+    outgoing::expand(&input)
+        .unwrap_or_else(syn::Error::into_compile_error)
+        .into()
 }
 
 /// Parses the optional `#[message(headers(Type))]` helper attribute of `#[derive(Message)]`.
@@ -383,18 +430,22 @@ pub fn derive_from_ref(item: TokenStream) -> TokenStream {
 /// [`NAME`](../ruststream/runtime/trait.OutSlot.html#associatedconstant.NAME) is the struct's
 /// identifier, used by startup errors and test assertions.
 ///
-/// The optional `#[publishes(Type = "channel", ..)]` attribute declares the slot's publish
-/// dictionary: each pair maps one message type to the channel it publishes to, enabling the
-/// typed publish path (`out.publish_typed(&value)`) and the `AsyncAPI` `send` operations.
-/// Several types may share a channel; listing one type twice is an error (its destination
-/// would be ambiguous). A marker without the attribute keeps the byte-level path only.
+/// The optional `#[publishes(Type, ..)]` attribute lists the message types the slot may
+/// publish, which is what the generated document reports for a handler leaving its `Out`
+/// parameter unrestricted. Each listed type declares its own destination with
+/// `#[derive(Outgoing)]`; listing one type twice is an error.
+///
+/// The name-carrying form `#[publishes(Type = "channel", ..)]` is deprecated: it maps the type
+/// to a channel per slot, which is what moved onto the message type. It keeps working (and keeps
+/// feeding the deprecated `out.publish_typed(&value)` path) until the deprecated surface is
+/// removed.
 ///
 /// ```ignore
 /// #[derive(OutSlot)]
 /// struct Encoded;
 ///
 /// #[derive(OutSlot)]
-/// #[publishes(ChunkDone = "chunks.done", Progress = "chunks.progress")]
+/// #[publishes(ChunkDone, Progress)]
 /// struct Events;
 ///
 /// #[subscriber("chunks", raw)]
@@ -437,50 +488,62 @@ pub fn derive_out_slot(item: TokenStream) -> TokenStream {
     let outgoing = if dictionary.is_empty() {
         quote!()
     } else {
+        // A listed type declaring its own destination contributes what it declares; one carrying
+        // the deprecated `= "channel"` form is described from the marker, as before.
         let entries = dictionary.iter().map(|entry| {
             let ty = &entry.ty;
-            let channel = &entry.channel;
+            let Some(channel) = &entry.channel else {
+                return quote! {
+                    __rs_entries.extend(
+                        <#ty as ::ruststream::runtime::OutMessages<#name>>::outgoing(),
+                    );
+                };
+            };
             quote! {
-                ::ruststream::runtime::OutgoingMessageMetadata::new(
-                    #channel,
-                    ::core::any::type_name::<#ty>(),
-                )
-                .with_message_name({
-                    #[allow(unused_imports)]
-                    use ::ruststream::__private::NoMessageProbe as _;
-                    ::ruststream::__private::Probe::<#ty>::new().message_name()
-                })
-                .with_message_description({
-                    #[allow(unused_imports)]
-                    use ::ruststream::__private::NoMessageProbe as _;
-                    ::ruststream::__private::Probe::<#ty>::new().message_description()
-                })
-                .with_payload_schema({
-                    #[allow(unused_imports)]
-                    use ::ruststream::__private::NoSchemaProbe as _;
-                    ::ruststream::__private::Probe::<#ty>::new().schema_json()
-                })
-                .with_headers_schema({
-                    #[allow(unused_imports)]
-                    use ::ruststream::__private::NoHeadersSchemaProbe as _;
-                    ::ruststream::__private::Probe::<#ty>::new().headers_schema_json()
-                })
+                __rs_entries.push(
+                    ::ruststream::runtime::OutgoingMessageMetadata::new(
+                        #channel,
+                        ::core::any::type_name::<#ty>(),
+                    )
+                    .with_message_name({
+                        #[allow(unused_imports)]
+                        use ::ruststream::__private::NoMessageProbe as _;
+                        ::ruststream::__private::Probe::<#ty>::new().message_name()
+                    })
+                    .with_message_description({
+                        #[allow(unused_imports)]
+                        use ::ruststream::__private::NoMessageProbe as _;
+                        ::ruststream::__private::Probe::<#ty>::new().message_description()
+                    })
+                    .with_payload_schema({
+                        #[allow(unused_imports)]
+                        use ::ruststream::__private::NoSchemaProbe as _;
+                        ::ruststream::__private::Probe::<#ty>::new().schema_json()
+                    })
+                    .with_headers_schema({
+                        #[allow(unused_imports)]
+                        use ::ruststream::__private::NoHeadersSchemaProbe as _;
+                        ::ruststream::__private::Probe::<#ty>::new().headers_schema_json()
+                    }),
+                );
             }
         });
         quote! {
             fn outgoing() -> ::std::vec::Vec<::ruststream::runtime::OutgoingMessageMetadata> {
-                ::std::vec![#(#entries),*]
+                let mut __rs_entries = ::std::vec::Vec::new();
+                #(#entries)*
+                __rs_entries
             }
         }
     };
-    let channels = dictionary.iter().map(|entry| {
+    let channels = dictionary.iter().filter_map(|entry| {
         let ty = &entry.ty;
-        let channel = &entry.channel;
-        quote! {
+        let channel = entry.channel.as_ref()?;
+        Some(quote! {
             impl ::ruststream::runtime::OutMessage<#name> for #ty {
                 const CHANNEL: &'static str = #channel;
             }
-        }
+        })
     });
 
     quote! {
@@ -585,6 +648,7 @@ pub fn derive_out_messages(item: TokenStream) -> TokenStream {
     quote! {
         #(#memberships)*
 
+        #[allow(deprecated)]
         impl<__RsM: ::ruststream::runtime::OutSlot> ::ruststream::runtime::OutMessages<__RsM>
             for #name
         where
@@ -634,19 +698,26 @@ fn out_messages_models(data: &syn::DataEnum) -> syn::Result<Vec<&Type>> {
     Ok(models)
 }
 
-/// One `Type = "channel"` pair of the `#[publishes(..)]` attribute.
+/// One entry of the `#[publishes(..)]` attribute: a type the slot may publish, optionally
+/// carrying the deprecated `= "channel"` destination.
 struct PublishesEntry {
     ty: Type,
-    /// The channel value: a string literal, or a const-evaluable `&'static str` expression (an
-    /// associated const initializer cannot read a `static`).
-    channel: Expr,
+    /// The channel value of the deprecated name-carrying form: a string literal, or a
+    /// const-evaluable `&'static str` expression (an associated const initializer cannot read a
+    /// `static`). `None` for the plain `#[publishes(Type, ..)]` list, where the destination
+    /// comes from the message type's own `#[derive(Outgoing)]` declaration.
+    channel: Option<Expr>,
 }
 
 impl Parse for PublishesEntry {
     fn parse(input: ParseStream) -> syn::Result<Self> {
         let ty: Type = input.parse()?;
-        input.parse::<Token![=]>()?;
-        let channel: Expr = input.parse()?;
+        let channel = if input.peek(Token![=]) {
+            input.parse::<Token![=]>()?;
+            Some(input.parse::<Expr>()?)
+        } else {
+            None
+        };
         Ok(Self { ty, channel })
     }
 }
@@ -669,8 +740,9 @@ fn publishes_dictionary(attrs: &[Attribute]) -> syn::Result<Vec<PublishesEntry>>
             {
                 return Err(syn::Error::new_spanned(
                     &entry.ty,
-                    "this type is already in the slot's publish dictionary: one type maps to \
-                     one channel per slot (use a second slot for a second destination)",
+                    "this type is already listed on the slot: a type appears once (with the \
+                     deprecated `= \"channel\"` form, a second entry would also make its \
+                     destination ambiguous)",
                 ));
             }
             dictionary.push(entry);
--- a/crates/ruststream-macros/src/lib.rs
+++ b/crates/ruststream-macros/src/lib.rs
@@ -15,7 +15,7 @@ use proc_macro2::TokenStream as TokenStream2;
 use quote::{ToTokens, quote};
 use syn::parse::{Parse, ParseStream};
 use syn::punctuated::Punctuated;
-use syn::{Attribute, DeriveInput, Expr, ItemFn, Meta, Token, Type, parse_macro_input};
+use syn::{Attribute, DeriveInput, Expr, Ident, ItemFn, Meta, Token, Type, parse_macro_input};
 
 use parse::{SubscriberArgs, doc_description};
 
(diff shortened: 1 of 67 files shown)


powersemmi (commented):
## Review: what the diff cannot carry

Pushed to this branch: one commit extending the `OutMessages` `on_unimplemented` note so it names the deprecated dictionary route as well. The case that actually hits that bound is a `#[derive(Message)]` type used as the third `Out` argument, whose `OutMessages` impl is conditional on `Self: OutMessage<M>`; the note listed only the three new declaration forms, so the reader was told to change the type when listing it on the marker is equally valid while the deprecated form lives. Snapshot regenerated; `just check`, `just test`, `--doc --no-default-features` and rustdoc with `-D warnings` all pass.

Everything below is reasoning, not diff commentary.

### Invariants the compiler is not holding

**A named slot's `#[publishes(..)]` list no longer bounds what the slot publishes.** Detailed inline on `TypedSlot::message`. `publish_typed` required `T: OutMessage<M>` on top of the parameter's set; the builder requires only the parameter's set, and the unrestricted parameter admits everything. What breaks if someone ignores it: a handler declared `Out<impl Publisher, Events>` publishes a `#[derive(Outgoing)]` type that `Events` never listed, it compiles, it runs, and the generated AsyncAPI document - still built from `<Events as OutSlot>::outgoing()` - does not mention the channel. Nothing fails, the document is just wrong, and it stays wrong until someone reads it against the code. This is the one finding I would not merge without a decision either way, because it is a compile-time guarantee that existed before this change and does not after it.

**A `CallerName` message contributes nothing to the document, by design, and that is now the easy path.** `.to(computed)` on a declaration-free type is exactly the byte path's old blind spot moved into the typed path: it is checked, encoded and contract-aware, but `OutMessages::outgoing()` returns an empty vec for it, so it is undocumentable by construction. That is what #231 decided, and the derive-with-no-`name` form is genuinely useful. The thing to watch is that it is now *shorter to write* than the fixed form (no attribute at all), so the incentive points at the undocumented case. If the document coverage ever regresses in a service, this is where it went.

**`#[derive(Outgoing)]` and `#[derive(Message)]` on one type are mutually exclusive by impl conflict, not by diagnostic.** Both emit `Message`, `MessageHeaders`, `ContainsMessage` and `OutMessages`, so writing both yields four E0119s naming the traits rather than one message naming the mistake. The rustdoc says not to; nothing enforces it more cheaply than the coherence error. Acceptable, but it is convention plus a wall of errors, not a designed diagnostic.

**A slot's testing attribution rides on `SlotPublisher::publish`, and the builder happens to preserve it.** `TypedSlot::message` sinks into `&self.slot`, whose `PublishSink` impl calls `P::publish`, which is where `record_slot_publish` lives - so `TestApp` assertions keep seeing builder publishes. That is load-bearing and invisible: any future sink that bypasses `Publisher::publish` (a direct broker call, a batching sink) silently drops slot attribution from the test harness, with green tests as the failure mode.

### Places that resist mental simulation

**Five type parameters, three of which are typestates with different vocabularies.** `Publish<Sink, Body, Enc, Hdrs, Dest>` mixes a resource (`Sink`), a payload discriminant (`Body`), a resolved value (`Enc`) and two state machines (`Hdrs`, `Dest`) in one tuple, and the reader has to hold which of the five each `impl` block is narrowing. It is correct and I have not touched it, but if it grows a sixth position, splitting the state machines out of the resource positions is the move.

**The `ErasedSink` / `ErasedPublishError` pair exists to reach one call site.** The deferred-retry fallback is the only consumer, and it is a wrapper plus an error newtype plus a `PublishSink` impl to route one republish through the builder. Consistency is a real reason and I would keep it, but it is an abstraction with exactly one caller and should be recognised as such when the deprecated pair is eventually deleted - at that point the wrapper is the only thing left holding `ErasedPublisher` open.

### Knowingly approximated or left undone

- `#[diagnostic::on_unimplemented]` on `ResolvedName` and `SatisfiesContract` does not render, because the bounds sit on inherent methods; the E0599 form carries the information instead. Inline detail on the trait.
- A templated address costs N+1 allocations, not the one the issue budgeted. Inline detail on the setter, with the trade that would remove them.
- `publish_erased` is a new required method on a public non-sealed trait, stamped as a patch-level change. Inline detail; this needs a call on the release classification.
- `#[publishes(A = "chan", B)]` mixes the deprecated and new forms in one attribute. Deliberate, for incremental migration; noted so the removal is understood as a parse change rather than a code path.

### Cross-check against #237

The two branches conflict only in `src/prelude.rs` and `src/runtime/mod.rs`, and only in the export lists - no duplicated machinery. The one shared file with real content, `crates/ruststream-macros/src/expand.rs`, merges cleanly: this PR touches `outgoing_method` and `slot_bounds` (the `OutMessage` -> `OutMessages` move), #237 restructures the attribute parsing and the definition emission, and the two do not overlap. The `#[derive(Outgoing)]` name-template machinery and #237's subscriber step traits are both typestates over a builder, but they share no code and should not: one binds address segments, the other closes settings slots, and the only thing worth keeping in step is that both name the missing piece in the type so the error can say which one it is. They do.

src/runtime/publish/builder.rs
```diff
@@ -0,0 +1,478 @@
...
+///
+/// This is what separates a publish that is ready to run from one whose call site has not said
+/// where the message goes. [`CallerName`] and [`NameTemplate`] deliberately do not implement it.
+#[diagnostic::on_unimplemented(
+    message = "this publish has no destination yet",
+    note = "a message type declaring no name takes one at the call: `.to(\"orders.archived\")`; \
+            one declaring a name template opens its placeholder setters with `.to()`, and the \
+            publish appears once every placeholder is bound"
+)]
+pub trait ResolvedName {
```
powersemmi: Worth stating why the `on_unimplemented` block here (and on `SatisfiesContract`) reads as decoration in the snapshots: rustc only renders `#[diagnostic::on_unimplemented]` when the failing obligation is a trait bound it is *solving*, and the bounds on `publish()` are bounds on an **inherent** method, so the failure surfaces as E0599 ("the method exists but its trait bounds were not satisfied") and the note is dropped. The snapshots are still readable because the E0599 body names the unsatisfied bound and the concrete type (`CallerName: ResolvedName`, `HeadersUnset: SatisfiesContract<WithHeaders<DoneMeta>>`), which is why the marker types are named after the mistake rather than after the machinery. Keeping the attributes costs nothing and starts working the day the bound moves onto a trait method, but a reader comparing them against `tests/ui/publish_no_destination.stderr` will otherwise assume the snapshot is stale.

src/runtime/publish/builder.rs
```diff
@@ -0,0 +1,478 @@
...
+    Hdrs: PublishHeaders + SatisfiesContract<T::Contract> + Send,
+{
+    type Error = Sink::Error;
+
+    async fn publish_at(self, name: &str) -> Result<(), PublishError<Sink::Error>> {
+        encode_and_send(self.sink, self.codec, self.body.0, self.headers, name).await
+    }
+}
+
+impl<Sink, Hdrs, Dest> Publish<Sink, RawBody<'_>, (), Hdrs, Dest>
```
powersemmi: The generic `Dest: ResolvedName` here is deliberate rather than sloppy, and it is worth saying so, because the obvious tightening makes the diagnostics worse. `RawBody` only ever starts at `CallerName` and only ever reaches `SuppliedName`, so `Publish<_, RawBody, _, _, FixedName>` is unconstructible and the `resolve("")` below is dead for that arm - the narrower `impl .. Publish<Sink, RawBody<'_>, (), Hdrs, SuppliedName<'_>>` would express exactly the reachable set and delete the empty-string argument. What it also deletes is the error for `raw(&bytes).publish()`: with the bound in place the call site gets E0599 naming `CallerName: ResolvedName`; with the impl narrowed it degrades to a bare "no method named `publish` found", because there is no unsatisfied bound left to report. Trading one unreachable arm for the diagnostic is the right way round, but the `""` is what pays for it.

src/runtime/publish/builder.rs
```diff
@@ -0,0 +1,478 @@
...
+}
+
+/// Sends one message, applying the headers position and the sink.
+async fn deliver<Sink: PublishSink, Hdrs: PublishHeaders>(
+    mut sink: Sink,
+    name: &str,
+    payload: &[u8],
+    headers: Hdrs,
+) -> Result<(), PublishError<Sink::Error>> {
+    let mut map = Headers::new();
```
powersemmi: Unconditionally building a map and calling `with_headers` on the outgoing message looks like a cost the old `publish_typed` path did not pay - it is not. `Headers::new()` is `HashMap::default()`, which does not allocate until the first insert, so `HeadersUnset` leaves the fixed-name path with exactly the allocations it had before (the codec's encode, and nothing else). `MapHeaders` then moves the caller's map in rather than copying it. Only `TypedHeaders` allocates, and only where the contract demanded a value anyway.

src/runtime/publish/ext.rs
```diff
@@ -0,0 +1,78 @@
...
+    {
+        raw_of(self, payload)
+    }
+
+    /// Starts a typed publish of a `#[derive(Outgoing)]` value, encoded with the crate's
+    /// default codec (name another one with `with_codec(..)`).
+    ///
+    /// Available when a codec feature is enabled; with none, every publish names its codec.
+    #[cfg(any(feature = "json", feature = "cbor", feature = "msgpack"))]
+    fn message<'a, T>(
```
powersemmi: The choice to put the crate default in the *call* position (`CallCodec(DefaultCodec::default())`) rather than adding a fourth `PublishCodec` impl for "nothing named one" is what keeps the ladder a single type parameter. A bare publisher has no level of its own, so the bottom of the ladder is the only level it can occupy, and `with_codec(..)` overwrites the whole position instead of having to outrank a marker. The consequence to keep in mind: on this surface `with_codec(..)` is not "more specific than the default", it is the same slot filled twice, so if a future surface ever wants to distinguish the two it needs a real level, not a flag.

src/runtime/slot.rs
```diff
@@ -511,6 +525,75 @@ where
...
+    ///     HandlerResult::Ack
+    /// }
+    /// # }
+    /// ```
+    pub fn message<'a, T, Index>(
+        &'a self,
+        value: &'a T,
+    ) -> Publish<&'a P, MessageBody<'a, T>, &'a EncodeCodec, HeadersUnset, T::Form>
+    where
+        Body: ContainsMessage<T, Index>,
```
powersemmi: This is the one place where the builder is strictly weaker than what it replaces, and it deserves a decision rather than silence. The deprecated `publish_typed` carried two bounds: `Body: ContainsMessage<T, Index>` (the *parameter's* declared set) **and** `T: OutMessage<M>` (membership in the *marker's* `#[publishes(..)]` dictionary). Only the first survives here. For `Out<impl Publisher, Events, (A, B)>` that is harmless, because the parameter names the set. For the unrestricted `Out<impl Publisher, Events>` it is not: `impl<T> ContainsMessage<T, Unrestricted> for ()` admits every `T`, so a handler can now `message(&Anything)` through a named slot whose `#[publishes(..)]` list never mentions it - while `outgoing_method` still documents that slot from `<Events as OutSlot>::outgoing()`, i.e. from the list. The generated document and the code can drift apart in a way that used to be a compile error. #231 says the dictionary "reduces to the list of types a slot may publish", which reads as *keeping* the restriction rather than dropping it. Restoring it needs a membership bound the marker can carry (an `M: SlotPublishes<T>` emitted per `#[publishes(..)]` entry, with a blanket impl for `DefaultSlot` so the unnamed slot stays open); that is a design call, so flagging it here rather than changing it.

src/runtime/publisher_registry.rs
```diff
@@ -6,20 +6,41 @@
...
 ///
 /// Blanket-implemented for every [`Publisher`], so any broker's publisher can be held as
 /// `Arc<dyn ErasedPublisher>` for the `retry_after` fallback.
 pub trait ErasedPublisher: Send + Sync {
+    /// Publishes one message, with whatever destination, payload and headers it carries.
+    ///
+    /// # Errors
+    ///
+    /// Returns the underlying publisher's error, boxed, if the broker rejects the publish.
+    fn publish_erased<'a>(
```
powersemmi: `publish_erased` is a **required** method added to a public, non-sealed trait, which the repo's pre-1.0 SemVer rules classify as breaking - while the deprecations in this same file are stamped `since = "0.6.4"`, i.e. a patch. The escape hatch the issue relies on ("blanket-implemented for every `Publisher`, so nothing outside implements it") holds for anything that *is* a publisher, but a foreign type that implements `ErasedPublisher` directly without implementing `Publisher` is legal today and stops compiling. It cannot be given a default body either: decomposing the `OutgoingMessage` back into name/payload/headers would have to call the deprecated `publish_message`, and there is no `Self: Publisher` supertrait to delegate through. So the options are to accept it as a knowing break (and release this as a minor, adjusting the `since` values), or to make the trait sealed / `#[doc(hidden)]` so the classification stops applying. Either way it should be a stated choice, not a side effect of the refactor.

src/schema.rs
```diff
@@ -100,11 +100,90 @@ pub trait HeadersContract: sealed::Sealed {}
...
+            the destination, `#[outgoing(name = \"orders.done\")]` fixes it, and \
+            `#[outgoing(name = \"orders.{{tenant}}.v1\")]` opens one setter per placeholder"
+)]
+pub trait OutgoingDestination {
+    /// The declared form: [`FixedName`], [`NameTemplate`] or [`CallerName`].
+    type Form: DestinationForm;
+
+    /// The declared address: the literal name, the template with its placeholders in braces, or
+    /// the empty string when the type declares none.
+    const ADDRESS: &'static str = "";
```
powersemmi: The `= ""` default on `ADDRESS` is what lets `CallerName` types implement the trait without a name, and it is the reason `ResolvedName::resolve` takes the declared address as an argument instead of reading it from the impl: `SuppliedName` ignores what it is handed, `FixedName` returns it, and nothing else can reach `publish()`. An alternative was to move the address onto the form (`FixedName(&'static str)`), which would have made the empty default unrepresentable - it was not taken because the form has to be `Default`-constructible from `message_of` before any value exists, and a form carrying data cannot be. Worth recording, since "an associated const that is meaningless for two of the three forms" is otherwise the first thing a reader flags.

crates/ruststream-macros/src/outgoing.rs
```diff
@@ -0,0 +1,495 @@
...
+        })
+        .collect();
+    let open = states
+        .iter()
+        .enumerate()
+        .filter(|(position, _)| *position != index)
+        .map(|(_, state)| quote!(#state));
+    let moved = fields.iter().enumerate().map(|(position, field)| {
+        if position == index {
+            quote!(#field: ::std::string::ToString::to_string(&value))
```
powersemmi: Each setter renders its segment into an owned `String` here, and `publish()` then renders the whole address again with `format!`, so a template with N placeholders costs N+1 allocations rather than the one the issue budgeted ("a templated address is rendered per publish"). The alternative is to let the bound state *be* the value type (`To<Cont, V0, V1>` with `V: Display`) and format once from the borrowed values. What that costs is the diagnostic: today the unbound state is `MissingSegment<Region>` against a required `String`, and the error names the segment; with a `Display` bound the same mistake reports "`MissingSegment<Region>: Display` is not satisfied", which is close but no longer says the terminal is missing. If the N extra allocations ever matter on this path, that is the trade to make, and the segment markers already carry the name needed to keep the message readable.

crates/ruststream-macros/src/lib.rs
```diff
@@ -614,19 +678,26 @@ fn out_messages_models(data: &syn::DataEnum) -> syn::Result<Vec<&Type>> {
...
+    /// comes from the message type's own `#[derive(Outgoing)]` declaration.
+    channel: Option<Expr>,
 }
 
 impl Parse for PublishesEntry {
     fn parse(input: ParseStream) -> syn::Result<Self> {
         let ty: Type = input.parse()?;
-        input.parse::<Token![=]>()?;
-        let channel: Expr = input.parse()?;
+        let channel = if input.peek(Token![=]) {
```
powersemmi: Making the channel optional per *entry* rather than per *attribute* means `#[publishes(A = "chan", B)]` parses, and the two halves are then documented by different mechanisms (`A` from the marker's dictionary, `B` from its own `#[derive(Outgoing)]`). That is deliberate - it is what lets an existing marker migrate one type at a time instead of in one commit - but it also means the duplicate check below is the only thing keeping a type from appearing in both forms at once. Worth knowing when the deprecated form is finally removed: the removal is `channel: Option<Expr>` becoming a hard parse error, not a second code path to delete.

PR #237, opened 2026-08-19
feat: express a subscriber's declarative settings through a builder the macro shares

### LINKED ISSUE #234: Express a subscriber's declarative settings through a builder the macro shares
## Problem

Everything declarative about a subscriber can only be said in one place, the attribute:
`#[subscriber("orders", workers(4), on_failure(decode = drop), start_at(..))]`. The attribute takes
any expression that stands on its own, so constants and build-time values work, but a setting the
service only knows while it wires itself up has no path at all. The one mount that could supply a
source at the call site, `include_on`, was never intended surface and has been removed.

The macro is also the only implementation of those settings. Each clause is parsed and expanded
into the definition's methods, so a new setting is added in the macro and nowhere else, and a user
who wants to build a subscriber programmatically has nothing to build it with.

The attribute carries a second kind of information as well, and the two are mixed: `raw` and
`batch(..)` do not describe data, they describe the shape of the handler, which is already visible
in its signature.

## Proposed solution

Split the two kinds and give the declarative one a builder that the macro shares.

**Declarative settings become a builder on the generated definition.** Name, worker policy, failure
policies, start position: each is a value, so each can be given in the attribute, or through the
builder, or partly in each.

```rust
#[subscriber]
async fn audit(order: &Order) -> HandlerResult {
    HandlerResult::Ack
}

b.include(audit.name(subject).workers(nonzero!(4)));
```

**The attribute always fixes the subscription type.** A subscriber is not one thing: a subject and
a JetStream consumer are different types, as are a Redis stream, a pub/sub channel and a list, and
so is a transactional subscription. The macro has to name that type to generate the definition, and
the broker extension traits described below bind on it, so it cannot be deferred. What can be
deferred are the values that fill it.

```rust
#[subscriber("orders")]                              // the by-name source, value fixed here
#[subscriber(RedisStream::new("orders").group("w"))] // a named kind, spelled out in full
#[subscriber(RedisStream::new("orders"))]            // a named kind, the rest from the builder
#[subscriber(RedisStream)]                           // a named kind, everything from the builder
```

The third is the same kind carrying only what it requires to exist, with the settings that depend
on values the service computes arriving through the broker extension methods below, which transform
the source the attribute already built:

```rust
b.include(bill.group(format!("workers-{shard}")));
```

An attribute with nothing in it is the by-name source with its value left out, so it reads as the
shortest of these rather than as a fourth thing:

```rust
#[subscriber]                                        // the by-name source, named by the builder
async fn audit(order: &Order) -> HandlerResult {
    HandlerResult::Ack
}

b.include(audit.name(subject));
```

It works wherever the broker implements the `Subscribe` capability, which every broker crate in the
family does today: a name is mapped to the subscription kind that broker considers its default, and
the configuration a kind needs beyond the name is set once on the broker. Redis maps a name to a
stream consumed under `RedisBroker::default_group`, Kafka to a topic under
`KafkaBroker::default_group`, and a broker that has nothing to configure maps it directly. Naming
another kind, a pub/sub channel or a JetStream consumer, is what the explicit source forms are for.

The last form works because a subscription kind is identified by a name and nothing else: every
source in the family is constructed from one string, `new(topic)` in Kafka, `new(subject)` in NATS,
`new(key)` or `new(channel)` in Redis. A small core trait says so, the source implements its own
constructor behind it, and the builder calls it when the name arrives, so nothing is ever built
from thin air and no source needs a `Default` that would make a stream without a key
representable.

Where a kind genuinely needs more than a name to exist, Pulsar's source takes a topic and a
subscription name, it does not implement the trait and the form does not compile for it. That is
the honest boundary: the kinds that can be named by one string offer it, the rest are written out.

The order in a chain follows from this: the name comes first because it constructs the source, the
broker settings then transform it, and the buffer wraps it last.

**The builder may only fill what the attribute left out.** A setting named in the attribute is
fixed in the definition's type and its builder method does not exist, so the two can never disagree
and there is no precedence rule to remember:

```rust
#[subscriber("orders", workers(4))]
async fn handle(order: &Order) -> HandlerResult {
    HandlerResult::Ack
}

b.include(handle.name("other"));    // does not compile: the name is already given
b.include(handle.on_failure(..));   // fine: the attribute said nothing about failures
```

An unnamed definition is not mountable until it is named, by the same mechanism: the definition
without a name lacks the piece `include` needs, so `b.include(audit)` is a compile error that names
the fix, the way an unbound `Out` slot already does.

**The macro expands into the same builder.** `#[subscriber("orders", workers(4))]` produces the
definition and then the calls a user would write by hand, so the settings have one implementation
and the attribute is sugar over it rather than a parallel path.

**Shape stays with the macro, and is inferred rather than declared.** Single message, batch, raw
bytes, the extractor parameters, the `Out` slots, the `Seek` parameter and the reply type all
follow from the signature, which only the macro can read. Since the shape is visible there, the
`raw` and `batch(..)` clauses stop being written: a handler taking `&[u8]` is raw, one taking
`&[T]` is a batch, and one taking `&[&[u8]]` is a batch of raw payloads.

**Broker settings reach the builder through the broker's own vocabulary.** Core cannot know that a
subscription has a JetStream stream or a durable consumer name, so it exposes one hook, a transform
over the source it is building, and a broker crate layers its own trait on top:

```rust
// in the broker crate
pub trait NatsSubscriber {
    fn jetstream(self, stream: impl Into<String>) -> Self;
    fn durable(self, name: impl Into<String>) -> Self;
}

impl<Def, W, F, P> NatsSubscriber for SubscriberBuilder<Def, SubscribeOptions, (W, F, P)> {
    fn jetstream(self, stream: impl Into<String>) -> Self {
        self.map_source(|source| source.jetstream(stream))
    }
    // ..
}
```

The bound names the source the builder currently holds, not the definition's associated source
type. A builder step may replace the source (a name resolved into a descriptor, a buffer wrapping
it), so the definition's own type stops describing what `map_source` will be handed; binding it
instead would leave the methods unavailable exactly where a step has been used. The settings state
stays generic, so the broker's vocabulary composes in any order with the core settings.

```rust
// at the call site, one chain, core settings and broker settings side by side
b.include(handle.workers(nonzero!(4)).jetstream("ORDERS").durable("orders-worker"));
```

The trait is local to the broker crate, so the orphan rule is satisfied, and the bound on the
source type means the methods simply do not exist on a builder for another broker. This is the
extension shape the `Out` slot dictionary already uses, where a broker adds its own trait over the
publisher rather than the core growing a vocabulary it cannot know. The user imports the broker's
trait to reach the methods, as with any extension trait.

**The batch shape belongs to the handler, the source of the batches to the mount.** A signature
taking `&[T]` says the handler wants several messages at once; whether they arrive that way is a
property of the broker, so it is settled where the definition is mounted, not where it is written.
The by-name source carries the broker's own subscriber type, so the check resolves per broker: the
same definition mounts natively where that kind batches and asks for the buffer where it does not,
and one definition mounted into two brokers can differ between the two mounts.

```rust
#[subscriber]
async fn correlate(events: &[Event]) -> HandlerResult {
    HandlerResult::Ack
}

// where the broker's by-name kind batches natively
b.include(correlate.name("events"));

// where it does not, the compiler asks for the buffer and the mount supplies it
b.include(correlate.name("events").buffered(nonzero!(10), Duration::from_millis(50)));
```

**The framework's own buffer is a separate, explicitly named setting.** It exists for brokers with
no batching of their own, where a handler still needs several messages at once because one delivery
only makes sense against another. Batches therefore come either from the broker, configured by the
broker methods above, or from the buffering adapter core wraps around a plain subscriber. Those are
different things, so the core setting says which one it is:

```rust
b.include(bill.name("orders").buffered(nonzero!(64), Duration::from_millis(10)));
```

Today that wrap is spelled inside the batch clause, with a turbofish so the macro can recover the
wrapped type: `#[subscriber(batch(Buffered::<Name>::new(Name::new("events")).max_size(nonzero!(2))))]`.
With the shape inferred from the signature the clause goes away, leaving the wrap as either the
attribute's source type or this builder step.

Which of the two applies is decided by the trait, at the mount site, where trait resolution
happens: the batch mount already requires the source's subscriber to implement `BatchSubscriber`,
brokers that batch on the wire implement it, and `Buffered<S>` is what supplies it for those that
do not. That direction is enforced. The opposite one, refusing a buffer over a subscriber that
already batches, would need a negative bound and is not expressible, so it stays legal and the
naming carries the meaning.

`BatchSubscriber` has no `#[diagnostic::on_unimplemented]` today, so mounting a batch handler on a
subscription that does not batch reports an unsatisfied bound and leaves the reader to work out the
fix. It should name it: the subscription does not batch, add the buffer.

One consequence to keep in mind: buffering changes the subscription type, from `S` to
`Buffered<S>`, so the builder step returns a builder over the wrapped source and the broker
extension methods, bound to the unwrapped type, stop applying past it. Broker settings come first
and the wrap last; the compiler enforces the order, and the diagnostic should say why rather than
reporting a missing method.

Naming it after the adapter keeps the trap shut: a name like `max_size` on the builder would read
as configuring the broker's batching, and would silently wrap a natively batching subscriber in a
second layer of buffering. With the setting named for what it does, that combination is still
expressible and is obviously a wrap.

**Raw and batch become a valid combination.** A raw batch is the typed batch without the decode
step, so the handler takes the payloads as a slice of byte slices borrowed from the batch's
messages, which the dispatcher holds for the duration of the call. Nothing is copied, and the
settlement rules are the ones the batch path already has.

**One mount entry point is left.** `include` and `include_batch` already dispatch identically, both
resolving the definition's form token, so the second name only ever said in prose what the
signature says in types. Once the shape is read off the signature there is nothing left for it to
distinguish, and `include` mounts every form. Removing it is a public removal on both the scope and
the router.

## Alternatives considered

- Let the builder override what the attribute set, with the builder winning. It makes a definition
  read differently depending on where it is mounted, and it needs a precedence rule where the type
  system can simply forbid the clash.
- Check at startup that a name was supplied. The mistake then surfaces wherever the service is
  first run rather than at the author's compiler, and a misuse the compiler can reject is rejected
  there.
- Keep `include_on(source, def)`. It made the source a property of the mount site; a broker source
  with its own builder chain belongs in the attribute, where it reads as part of the declaration.
- Expose the builder only, and drop the attribute. The attribute is what makes a subscriber
  readable at a glance, and the signature analysis behind it cannot be done anywhere else.

## Additional context

Additive. An attribute that names every setting behaves exactly as it does today; the builder is
reachable only for settings the attribute left out, and the shape clauses keep parsing while they
are being retired.

Two things to watch when the work starts. The macro currently rejects illegal combinations with
messages of its own, snapshot-tested under `tests/ui`; anything that moves into the builder has to
keep that quality through the type system, with `#[diagnostic::on_unimplemented]` where a bound is
what fails. And the generated document reads a subscriber's name from the mounted registration, so
a name supplied through the builder reaches it unchanged.

powersemmi:
# Description

Everything declarative about a subscriber could be said in one place only, the attribute, so a
setting the service knows only while it wires itself up had no path in at all. And the macro was
the only implementation of those settings, which left a user who wants to build a subscriber
programmatically with nothing to build it with.

The declarative settings - name, worker policy, failure policies, start position, the framework's
buffer - now live on a builder over the generated definition, which both registration surfaces
accept wherever they accept a definition. The attribute expands into that same builder, so a
setting has one implementation and the attribute is sugar over it rather than a parallel path.

Which settings a builder still accepts is carried in its type: one step trait per setting,
implemented only for the state where that setting is still open. There is no precedence rule to
remember - naming a setting twice does not compile, and each step carries its own message saying
so. A definition whose subscription was never named is not mountable by the same mechanism.

The attribute always fixes the subscription *kind*, in one of four forms from an empty attribute
up to a fully spelled-out source; what may be deferred is the value that fills it. The two short
forms rest on a small core trait giving a kind its own from-name constructor, so nothing is ever
built from thin air, and a kind that genuinely needs more than a name simply does not offer it.
Broker-specific settings arrive through a broker-owned trait layered over a core source-transform
hook, the same extension shape the `Out` slot vocabulary already uses.

The second kind of information the attribute carried - the shape of the handler - is read off the
signature, which is the only place it was ever really written: one message, one delivery's
payload, a whole decoded batch, or a batch of payloads. That last one is new: a raw batch is the
typed batch without the decode step, with the payloads borrowed from the batch's own messages.
Whether batches arrive at all is a property of the broker, so it is settled where the definition
is mounted, and the batching capability now says which fix applies when the subscription has no
batching of its own.

Fixes #234

## Type of change

## Checklist

      `examples/` where applicable

## Notes for the reviewer

Decisions the issue left open, taken here:

- **A subscription source is now `Clone`.** The mount rebuilds the source from the builder rather
  than re-evaluating the attribute's expression, so a broker descriptor used with `#[subscriber]`
  derives `Clone`. It is configuration, and cloning it is what lets one definition mount on two
  brokers; documented in the broker-authors guide.
- **The failure policies are one setting, not two.** An `on_failure(..)` clause naming only
  `decode` still closes the whole slot. Splitting the panic and materialization keys into separate
  slots would double the state without a case asking for it.
- **The materialization policy travels on the delivery context.** A `FromHeaders` contract is
  parsed in the handler body, so the macro used to interpolate the attribute's `decode` policy
  there; it now reads the effective policy off the context, which is what makes a policy named on
  the builder apply to the header contract as well.
- **A raw batch does not take `Out` / `Seek` yet.** The combination needs a parallel injection
  path at the byte shape; the macro rejects it with a message naming the alternatives rather than
  landing a half-shape.
- **The retiring clauses keep their meaning while they parse.** `batch(..)` still says "this
  slice is a batch", so a batch of `u8` stays expressible; without it a `&[u8]` parameter reads as
  the payload, which is what nobody would mean otherwise. `raw` with `batch(..)` now asks for
  `&[&[u8]]` instead of being rejected outright.
- **Double buffering stays legal.** Refusing a second wrap would need a negative bound, the same
  reason the issue gives for wrapping a natively batching subscription.

### DIFF
--- a/crates/ruststream-macros/src/expand.rs
+++ b/crates/ruststream-macros/src/expand.rs
@@ -14,74 +14,84 @@ use crate::parse::{
 };
 
 pub(crate) fn subscriber(args: &SubscriberArgs, func: &ItemFn) -> syn::Result<TokenStream> {
-    reject_raw_combinations(args, func)?;
+    reject_reply_combinations(args)?;
     let parts = handler_parts(args, func)?;
-    let body = if args.raw.is_some() {
-        // The remaining combinations are already rejected above; publish_raw(..) selects the
-        // byte reply form, otherwise raw is the plain form (injected when the signature
-        // carries Out / Seek parameters - the input axis makes raw compose with them).
-        match &args.publish_raw {
-            Some(reply_topic) => expand_publishing(&parts, func, reply_topic, true, true)?,
-            None if !parts.outs.is_empty() || parts.seek.is_some() => expand_injected(&parts, true),
-            None => expand_subscribing(&parts, true),
-        }
-    } else if let Some(reply_topic) = &args.publish_raw {
-        expand_publishing(&parts, func, reply_topic, true, false)?
-    } else {
-        match (&args.batch, &args.publish) {
-            (true, Some(reply_topic)) => expand_batch_publishing(&parts, func, reply_topic)?,
-            (true, None) if !parts.outs.is_empty() || parts.seek.is_some() => {
-                expand_batch_injected(&parts, func)
+    reject_shape_combinations(args, func, &parts)?;
+    let injected = !parts.outs.is_empty() || parts.seek.is_some();
+    let body = match parts.shape {
+        Shape::RawBatch => expand_raw_batch(&parts, func),
+        Shape::Batch => match &args.publish {
+            Some(reply_topic) => expand_batch_publishing(&parts, func, reply_topic)?,
+            None if injected => expand_batch_injected(&parts, func),
+            None => expand_batch(&parts, func),
+        },
+        // The input axis is a flag, not a form, so the byte input composes with every
+        // single-message form.
+        Shape::Single | Shape::Raw => {
+            let raw = parts.shape == Shape::Raw;
+            if let Some(reply_topic) = &args.publish_raw {
+                expand_publishing(&parts, func, reply_topic, true, raw)?
+            } else if let Some(reply_topic) = &args.publish {
+                expand_publishing(&parts, func, reply_topic, false, raw)?
+            } else if injected {
+                expand_injected(&parts, raw)
+            } else {
+                expand_subscribing(&parts, raw)
             }
-            (true, None) => expand_batch(&parts, func),
-            (false, Some(reply_topic)) => {
-                expand_publishing(&parts, func, reply_topic, false, false)?
-            }
-            (false, None) if !parts.outs.is_empty() || parts.seek.is_some() => {
-                expand_injected(&parts, false)
-            }
-            (false, None) => expand_subscribing(&parts, false),
         }
     };
     Ok(body.into())
 }
 
-/// Rejects the argument combinations the byte-level forms do not take: `raw` with batches (one
-/// delivery's bytes only) or the decode failure policy (there is no decode step), an encoded
-/// `publish(..)` reply off raw bytes (the reply of a raw handler is bytes - `publish_raw`), both
-/// reply clauses at once, and a batch with a byte reply.
-fn reject_raw_combinations(args: &SubscriberArgs, func: &ItemFn) -> syn::Result<()> {
+/// The combinations that are wrong before the signature is even read: two reply clauses at once.
+fn reject_reply_combinations(args: &SubscriberArgs) -> syn::Result<()> {
     if let (Some(_), Some(publish_raw)) = (&args.publish, &args.publish_raw) {
         return Err(Error::new_spanned(
             publish_raw,
             "publish(..) and publish_raw(..) are mutually exclusive: one reply, one destination",
         ));
     }
-    if let Some(publish_raw) = &args.publish_raw
-        && args.batch
-    {
+    Ok(())
+}
+
+/// The combinations the inferred shape rules out: an encoded reply off undecoded bytes (the reply
+/// of a raw handler is bytes), a byte reply off a batch, a decode policy where nothing is
+/// materialized, and the injection parameters the raw batch form does not carry yet.
+fn reject_shape_combinations(
+    args: &SubscriberArgs,
+    func: &ItemFn,
+    parts: &HandlerParts<'_>,
+) -> syn::Result<()> {
+    let raw_input = matches!(parts.shape, Shape::Raw | Shape::RawBatch);
+    let batched = matches!(parts.shape, Shape::Batch | Shape::RawBatch);
+    if raw_input && let Some(publish) = &args.publish {
         return Err(Error::new_spanned(
-            publish_raw,
-            "publish_raw(..) is not supported together with batch(..) yet; publish per message \
-             or use the encoded batch reply form",
-        ));
-    }
-    let Some(raw) = &args.raw else {
-        return Ok(());
-    };
-    if args.publish.is_some() {
-        return Err(Error::new(
-            raw.span(),
+            publish,
             "the reply of a raw handler is bytes and is never encoded; use \
              publish_raw(\"dest\") instead of publish(..)",
         ));
     }
-    if args.batch {
-        return Err(Error::new(
-            raw.span(),
-            "raw is not supported together with batch(..); a raw handler takes one delivery's \
-             payload as `&[u8]`",
+    if batched && let Some(publish_raw) = &args.publish_raw {
+        return Err(Error::new_spanned(
+            publish_raw,
+            "publish_raw(..) is not supported together with a batch handler yet; publish per \
+             message or use the encoded batch reply form",
         ));
+    }
+    if parts.shape == Shape::RawBatch {
+        if !parts.outs.is_empty() || parts.seek.is_some() {
+            return Err(Error::new_spanned(
+                &func.sig,
+                "a raw batch handler does not take Out / Seek parameters yet; take the batch as \
+                 `&[T]` for the decoded form, or the payload as `&[u8]` per delivery",
+            ));
+        }
+        if let Some(publish) = &args.publish {
+            return Err(Error::new_spanned(
+                publish,
+                "a raw batch handler has no reply form yet; publish from the body",
+            ));
+        }
     }
     // With a FromHeaders parameter the decode policy still has a job on a raw handler: it
     // settles a header contract that fails to parse.
@@ -90,14 +100,16 @@ fn reject_raw_combinations(args: &SubscriberArgs, func: &ItemFn) -> syn::Result<
         .inputs
         .iter()
         .any(|arg| matches!(arg, FnArg::Typed(pt) if from_headers_ty(&pt.ty).is_some()));
-    if let Some(failure) = &args.on_failure
+    if raw_input
+        && let Some(failure) = &args.on_failure
         && failure.decode.is_some()
         && !has_from_headers
     {
-        return Err(Error::new(
-            raw.span(),
-            "on_failure(decode = ..) does not apply to raw: the payload is not decoded and \
-             there is no FromHeaders parameter; keep only on_failure(panic = ..)",
+        return Err(Error::new_spanned(
+            func.sig.inputs.first(),
+            "on_failure(decode = ..) does not apply to an undecoded payload: this handler takes \
+             the bytes as delivered and declares no FromHeaders parameter; keep only \
+             on_failure(panic = ..)",
         ));
     }
     Ok(())
@@ -121,13 +133,54 @@ struct HandlerParts<'a> {
     extractors: Vec<(&'a Pat, &'a Type)>,
     outs: Vec<OutParam<'a>>,
     seek: Option<(&'a Pat, &'a Type)>,
-    workers_method: TokenStream2,
-    failure_method: TokenStream2,
-    /// The `FailurePolicy` value applied to `FromHeaders` extraction (the `decode` key).
-    headers_policy: TokenStream2,
+    /// What the handler consumes per invocation, inferred from its message parameter.
+    shape: Shape,
+    /// The builder calls the attribute's settings expand into, chained onto
+    /// `SubscriberBuilder::new(..)` inside the generated `Declared::declare`.
+    settings_chain: TokenStream2,
+    /// The source type the declared builder ends up over: the attribute's source, wrapped in
+    /// `StartAt<_, _>` when the attribute names a start position.
+    settings_source_ty: TokenStream2,
+    /// Which settings the attribute fixed, as the builder's `(workers, failures, position)`
+    /// state tuple.
+    settings_state_ty: TokenStream2,
     /// The `headers_schema()` def-method override lifted from the first `FromHeaders<T>`
     /// parameter's contract type, or empty when the handler takes none.
     headers_schema: TokenStream2,
+}
+
+/// The `Declared` impl every expansion emits next to its definition: the mount form, and the
+/// builder the attribute's settings expand into.
+///
+/// This is what makes the attribute sugar over the builder rather than a second implementation:
+/// `#[subscriber("orders", workers(4))]` produces exactly the calls a user would write at the
+/// mount site.
+fn declaration(parts: &HandlerParts<'_>, form: &TokenStream2) -> TokenStream2 {
+    let HandlerParts {
+        name,
+        source_expr,
+        settings_chain,
+        settings_source_ty,
+        settings_state_ty,
+        ..
+    } = parts;
+    quote! {
+        impl ::ruststream::runtime::Declared for #name {
+            type Form = #form;
+            type Settings = ::ruststream::runtime::SubscriberBuilder<
+                #name,
+                #settings_source_ty,
+                #settings_state_ty,
+            >;
+
+            fn declare(self) -> Self::Settings {
+                #[allow(unused_imports)]
+                use ::ruststream::runtime::SubscriberSettings as _;
+                ::ruststream::runtime::SubscriberBuilder::new(self, #source_expr)
+                    #settings_chain
+            }
+        }
+    }
 }
 
 /// The per-delivery context type the handler named in its `ctx: &mut Context<'_, C>` parameter,
@@ -439,22 +492,26 @@ fn extractor_where(
 /// extraction runs `reject` (a `return` settling the delivery by the rejection's `HandlerResult`).
 ///
 /// A `FromHeaders<T>` parameter takes the policy-aware path instead of the generic
-/// [`FromContext`] call, so the subscriber's `on_failure(decode = ..)` policy (interpolated as
-/// `headers_policy`) applies to the header contract exactly like it applies to the payload codec.
+/// [`FromContext`] call, and reads the policy off the delivery context rather than off the
+/// attribute: the effective materialization policy is the one the mount resolved, whether it was
+/// named in `on_failure(..)` or on the builder.
 fn extractor_prelude(
     extractors: &[(&Pat, &Type)],
     ctx_param: &TokenStream2,
     ctx_ty: &TokenStream2,
     state: &TokenStream2,
-    headers_policy: &TokenStream2,
     reject: &TokenStream2,
 ) -> TokenStream2 {
     let binds = extractors.iter().map(|(pat, ty)| {
         if from_headers_ty(ty).is_some() {
             return quote! {
-                let #pat = match <#ty>::extract(&mut *#ctx_param, #headers_policy) {
-                    ::core::result::Result::Ok(__rs_value) => __rs_value,
-                    ::core::result::Result::Err(__rs_err) => { #reject; }
+                let #pat = {
+                    // Read before the mutable borrow the extraction takes.
+                    let __rs_policy = #ctx_param.decode_policy();
+                    match <#ty>::extract(&mut *#ctx_param, __rs_policy) {
+                        ::core::result::Result::Ok(__rs_value) => __rs_value,
+                        ::core::result::Result::Err(__rs_err) => { #reject; }
+                    }
                 };
             };
         }
@@ -472,10 +529,10 @@ fn extractor_prelude(
     quote!(#(#binds)*)
 }
 
-/// Renders the `on_failure(..)` clause as an override of the def's defaulted `failure_policies`
-/// method, or nothing when the clause is absent. Only the keys named in the clause are set; the
-/// rest keep the runtime defaults.
-fn failure_method(args: &SubscriberArgs) -> TokenStream2 {
+/// Renders the `on_failure(..)` clause as the builder step it expands into, or nothing when the
+/// clause is absent. Only the keys named in the clause are set; the rest keep the runtime
+/// defaults.
+fn failure_step(args: &SubscriberArgs) -> TokenStream2 {
     let Some(failure) = &args.on_failure else {
         return quote!();
     };
@@ -490,9 +547,7 @@ fn failure_method(args: &SubscriberArgs) -> TokenStream2 {
         .map(failure_policy_tokens)
         .map(|policy| quote!(.with_decode(#policy)));
     quote! {
-        fn failure_policies(&self) -> ::ruststream::runtime::FailurePolicies {
-            ::ruststream::runtime::FailurePolicies::default() #panic #decode
-        }
+        .on_failure(::ruststream::runtime::FailurePolicies::default() #panic #decode)
     }
 }
 
@@ -505,11 +560,12 @@ fn failure_method(args: &SubscriberArgs) -> TokenStream2 {
 /// handler has no typed input to carry a contract, so without a `FromHeaders` parameter it
 /// emits nothing.
 fn headers_schema_method(
-    args: &SubscriberArgs,
+    shape: Shape,
     extractors: &[(&Pat, &Type)],
     input_ty: &Type,
 ) -> syn::Result<TokenStream2> {
-    if args.batch
+    let batch = matches!(shape, Shape::Batch | Shape::RawBatch);
+    if batch
         && let Some((_, ty)) = extractors
             .iter()
             .find(|(_, ty)| from_headers_ty(ty).is_some())
@@ -526,7 +582,7 @@ fn headers_schema_method(
         .iter()
         .find_map(|(_, ty)| from_headers_ty(ty))
         .map(|contract| {
-            if args.batch {
+            if batch {
                 vec_inner_ty(contract).unwrap_or(contract)
             } else {
                 contract
@@ -534,7 +590,7 @@ fn headers_schema_method(
         })
         .map_or_else(
             || {
-                if args.raw.is_some() {
+                if matches!(shape, Shape::Raw | Shape::RawBatch) {
                     return quote!();
                 }
                 quote! {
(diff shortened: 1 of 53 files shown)


src/runtime/settings/forward.rs
```diff
@@ -0,0 +1,283 @@
...
+use crate::runtime::publishing::{PublishingCall, PublishingDef};
+use crate::runtime::slot::{BindSlots, HasSlots};
+use crate::runtime::subscriber_def::SubscriberDef;
+
+/// The metadata methods every definition trait shares, forwarded to the wrapped definition, plus
+/// the two settings the builder owns.
+macro_rules! forward_common {
+    ($def:ty) => {
+        fn source(&self) -> Src {
+            self.source.clone()
```
powersemmi: This one line is where the cross-repo `Clone` requirement comes from, and it is worth being explicit that it is a choice, not a necessity. `SubscriberDef::source(&self)` hands out an owned source from a shared borrow; a generated definition is a unit struct that used to rebuild the source by re-evaluating the attribute's expression, and a builder holds one constructed value instead, so `&self` leaves cloning as the only option. It is avoidable: every mount reads the source exactly once, so a `fn into_parts(self) -> (Source, Handler, ..)` on the six definition traits would move it out and need no `Clone` at all. The comment on `SubscriberBuilder::into_parts` ("a step can wrap it without demanding `Clone` of a broker's descriptor") is true of the *steps* and reads, next to this, as if the type never needed it. What tipped the balance is that the definition traits are core-internal (the macro implements them; broker crates never do), whereas restructuring six of them plus every mount tail is a much larger change than deriving `Clone` on a configuration struct. That is a defensible trade, but it should be stated as one: the cost lands in other repositories, and the benefit is that this one stays small.

The part that was not defensible was *where* the requirement would be discovered. `SubscriptionSource` has no `Clone` supertrait and the conformance suite did not require it, so a broker crate could ship, pass conformance, and only fail in a user's `b.include(..)` with an unsatisfied `SubscriberBuilder<..>: SubscriberDef`. I have added `Clone` to `harness::lifecycle`'s source bound on this branch, so the broker's own conformance run is what reports it. Making it a supertrait of `SubscriptionSource` would be better still - the error would land at the impl - but that is a breaking change to a core trait and belongs to you, not to a review.

src/runtime/settings.rs
```diff
@@ -0,0 +1,526 @@
...
+            failures: FailurePolicies::default(),
+            _state: PhantomData,
+        }
+    }
+}
+
+impl<Def, Src, State> SubscriberBuilder<Def, Src, State> {
+    /// The pieces a step rebuilds from: the source moves out, so a step can wrap it without
+    /// demanding `Clone` of a broker's descriptor.
+    fn into_parts(self) -> (Def, Src, Workers, FailurePolicies) {
```
powersemmi: Worth noting what this does and does not buy. Moving the source out per step is what lets `buffered`, `start_at` and `map_source` *wrap* a source they cannot clone, so the chain itself is `Clone`-free - which is the useful half. The doc comment's "without demanding `Clone` of a broker's descriptor" reads as a property of the type, and the forwarding impls demand it anyway; keeping the sentence scoped to the steps would stop a reader concluding the requirement is not there.

src/runtime/settings.rs
```diff
@@ -0,0 +1,526 @@
...
+/// A setting the attribute left out, still fillable at the mount site.
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq, Hash)]
+pub struct Open;
+
+/// A setting already named, in the attribute or by an earlier builder call.
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq, Hash)]
+pub struct Fixed;
+
+/// The settings state of a definition whose attribute named none of them.
+pub type AllOpen = (Open, Open, Open);
```
powersemmi: The state being a bare 3-tuple rather than a named type is what makes each step trait implementable for one position (`(Open, F, P)`, `(W, Open, P)`, `(W, F, Open)`) without a projection, and it is why the steps are three flat impls instead of a type-level map. It has a visible cost: the definition type now renders as `SubscriberBuilder<handle, Name, (Open, Open, Open)>` in every mount diagnostic that used to say `handle` - see the refreshed `subscriber_out_unbound_slot` and `subscriber_publish_unserializable_reply` snapshots. Making `AllOpen` a struct would shorten the common case but breaks per-position matching, so the tuple stays. The mitigation that does work is the one already applied: the `on_unimplemented` headlines carry the message, so the long type only ever appears in the `help` lines below it.

src/runtime/settings.rs
```diff
@@ -0,0 +1,526 @@
...
+    /// definition, the definition itself for a hand-written one.
+    type Settings;
+
+    /// Applies the settings the declaration fixed.
+    fn declare(self) -> Self::Settings;
+}
+
+// A hand-written definition names its own form and carries its own settings, so it is already
+// what the mount machinery drives.
+impl<T: IncludeDef> Declared for T {
```
powersemmi: This blanket impl and the macro's own `Declared` impl are why the generated definition stopped implementing `IncludeDef`: the two would overlap, and coherence has no way to prefer one. The consequence is worth stating, because it is not obvious from either side - a hand-written definition keeps `Settings = Self` and therefore gets *none* of the builder steps, since `NameStep` and friends are implemented only for `SubscriberBuilder`. Calling `.workers(..)` on a hand-written def compiles into nothing useful today; it fails on `Self::Settings: WorkersStep`, which is the right answer but arrives as an unsatisfied bound on a type the author wrote by hand. If hand-written definitions should reach the builder later, the route is for them to return a `SubscriberBuilder` from their own `Declared` impl, exactly as the test stub in this file does.

src/runtime/context.rs
```diff
@@ -73,6 +73,10 @@ pub struct Context<'a, C = (), S = ()> {
     delivery: &'a Delivery,
     after: Vec<AfterHook>,
     failfast: Option<&'a ErrorShutdown>,
+    /// The subscriber's materialization policy, set by the dispatcher from the definition's
+    /// failure policies. It reaches the handler body because a `FromHeaders` contract is parsed
+    /// there rather than in the decode adapter.
+    decode: FailurePolicy,
```
powersemmi: Judged this one specifically because it lands on the per-delivery path. It is fine: `Context` is a stack value built per delivery, `FailurePolicy` is a `Copy` enum whose largest variant is `RetryAfter(Duration)`, so this adds one field to a struct that already exists and no allocation, no indirection and no clone. It is also not avoidable by moving the policy into a type parameter, because the whole point of the change is that the policy may arrive from the mount site as a runtime value.

What is held by convention rather than by the type is the *default*: `Context::new` seeds `FailurePolicy::Drop` and both dispatch entry points then overwrite it with `.with_decode_policy(..)`. A third dispatch path that forgets the call silently gets `Drop` for every `FromHeaders` contract, with no test failing - the value happens to match the documented default, which is what makes the omission invisible. Taking the policy as a `new()` parameter would make it unrepresentable; it was not done because `Context::new` has a long tail of call sites in tests. Worth a decision rather than an inherited default.

src/runtime/batch.rs
```diff
@@ -460,6 +476,55 @@ where
...
+        let subscription = ctx.name().to_owned();
+        if batch.is_empty() {
+            return;
+        }
+        #[cfg(feature = "otel")]
+        record_batch_size(&subscription, batch.len());
+        let tasks = ctx.tasks().clone();
+        // The views borrow the deliveries, so they are gone before the batch is settled.
+        let result = {
+            let payloads: Vec<&[u8]> = batch.iter().map(IncomingMessage::payload).collect();
```
powersemmi: The `Vec<&[u8]>` is one allocation per batch and is what the `&[&[u8]]` signature costs: a slice of slices has to exist somewhere, and the alternative shapes (an iterator parameter, or a handler called per element) are not the batch contract. Nothing is copied - the views borrow the deliveries, and the scope block is what guarantees they are dropped before `settle_batch` takes ownership. The `ctx.name().to_owned()` above happens before the empty-batch return, which is one wasted `String` on an empty batch; that is inherited from `TypedBatch` and consistent with it rather than new here.

crates/ruststream-macros/src/parse.rs
```diff
@@ -33,6 +33,43 @@ pub(crate) struct SubscriberArgs {
...
+    /// `#[subscriber(RedisStream)]`: a named kind, its value left to the mount site.
+    OpenKind(Type),
+    /// `#[subscriber("orders")]` / `#[subscriber(RedisStream::new("orders").group("w"))]`: the
+    /// value is fixed here too.
+    Fixed(Expr),
+}
+
+/// The clause keywords, so a leading one is not mistaken for a source expression: they parse as
+/// expressions too (`workers(4)` is a call, `raw` a path).
+const CLAUSES: &[&str] = &[
```
powersemmi: `CLAUSES` is a second, hand-maintained copy of the keyword set that the `Parse for SubscriberArgs` loop matches on, and nothing keeps the two in step. If a clause is added to the match without being added here, `#[subscriber(new_clause(..))]` is no longer recognised as a clause: `parse_source` takes it for a source expression, and the author gets `source_tokens`' "the macro follows the chain down to the base `Type::new(..)`" error about a clause they spelled correctly. The list cannot simply drive the match, because each clause parses a different payload, so the honest options are a macro that emits both from one place or a unit test asserting every matched keyword appears here. The `fork.is_empty() || peek(,) || peek(Paren)` discrimination itself is right and the `raw::Frames` case in the comment is the one that justifies it.

crates/ruststream-macros/src/expand.rs
```diff
@@ -910,40 +943,69 @@ fn split_seek<'a>(
...
-fn input_type<'a>(
+/// What the handler consumes per invocation, read off its message parameter rather than declared
+/// in the attribute: `&T` is one decoded message, `&[u8]` one delivery's payload as delivered,
+/// `&[T]` a whole decoded batch, and `&[&[u8]]` a batch of payloads.
+///
+/// A batch of `u8` values is not a thing anyone means, so `&[u8]` reads as the payload; spelling
+/// `batch(..)` in the attribute still says otherwise, which is what keeps the retiring clause
+/// meaning what it always did.
+#[derive(Clone, Copy, PartialEq, Eq)]
+enum Shape {
```
powersemmi: Replacing the `(raw: bool, batch: bool)` pair with this enum is what makes the four-way dispatch in `subscriber` exhaustive rather than a nest of `if`s, and it is what let `reject_raw_combinations` split into a pre-signature check and a post-signature one. One ordering decision inside `resolve_shape` is load-bearing and easy to misread: the `&[&[u8]]` test runs *before* the `raw` clause is consulted, so a handler taking that parameter is a raw batch whatever the attribute says, and the `raw` + `batch(..)` error below it can only fire for some other element type. That is deliberate - no second reading of `&[&[u8]]` exists - but it means the two branches are not alternatives, and someone reordering them for tidiness would turn the retiring `batch(..)` clause back into a rejection.

src/conformance/harness.rs
```diff
@@ -112,7 +113,7 @@ pub async fn lifecycle<B, MkBroker, Src, MkSrc, Pub, MkPub>(
 ) where
     B: Broker,
     MkBroker: Fn() -> B,
-    Src: SubscriptionSource<Connected<B>> + Send,
+    Src: SubscriptionSource<Connected<B>> + Clone + Send,
```
powersemmi: Added on this branch. The reasoning is on the `forward.rs` comment; in short, `Clone` on a broker's descriptor became part of the broker contract in this change, and the contract is what conformance exists to check. Without this bound the requirement is discovered in a user's `b.include(..)`, as an unsatisfied `SubscriberBuilder<..>: SubscriberDef`; with it, the broker's own conformance run names it. This does tighten a public signature, so it is a breaking change for a broker crate that has not derived `Clone` yet - which is precisely the set of crates the follow-up adoption wave has to visit anyway.

powersemmi (commented):
## Review: what the diff cannot carry

Pushed to this branch, one commit:

- **The red check.** `guides/subscribers.md` pointed at `../getting-started/index.md`, which does not exist; the strict build rejects it. The prelude it meant has no page on this site, so the link now goes to the API reference, which is what every other prelude-adjacent reference in `docs/` does.
- **The `Clone` contract, moved to where it is checked.** `harness::lifecycle`'s source bound now requires `Clone`. Reasoning inline; the short version is that a broker could pass conformance and still fail at a user's mount site.
- **Import hygiene** in the files this change already touches: `std::fmt::*` and `std::borrow::Cow` inlined in impl headers and signatures (`batch.rs`, `subscription.rs`, `router/builder.rs`, `memory/mod.rs`, `app/scope.rs`) hoisted, and the `settings.rs` test module's `crate::nonzero!` / `crate::SubscriptionSource::<..>` call-site paths turned into imports.

`just check`, `just test`, the trybuild suite under `RUN_UI_TESTS=1`, `--doc --no-default-features`, rustdoc with `-D warnings` and `lint_doc_fences.py` all pass. Everything below is reasoning, not diff commentary.

### The two decisions that reach beyond the issue

**`Clone` on broker descriptors: necessary as implemented, avoidable in principle.** Detailed inline on `forward.rs`. It falls out of keeping `SubscriberDef::source(&self)` as it is; moving the source out once (`into_parts(self)`) across the six definition traits would remove it entirely, and those traits are core-internal, so no broker would have to change. The trade taken is smaller code here for a requirement in every broker repository, and `Clone` on a configuration struct is genuinely cheap - but the cost is other people's crates and it should be recorded as a decision. The sharper problem was discoverability, which the pushed commit fixes.

**`Context::decode_policy()`: necessary, and free.** Detailed inline. One `Copy` enum field on a per-delivery stack value - no allocation, no indirection, no clone - and it cannot be a type parameter because the policy may be a runtime value from the mount site. I would not change it. What I would decide is the seeded `FailurePolicy::Drop` default, which is currently kept correct by both dispatch sites remembering to call `with_decode_policy`.

### Invariants the compiler is not holding

**The macro's clause list is duplicated.** `parse::CLAUSES` and the keyword match in `Parse for SubscriberArgs` have to agree, and nothing checks that they do. A clause added to one and not the other is parsed as a source expression, and the author gets a source-type error about a correctly spelled clause. Inline detail with the two ways out.

**`Context::new`'s decode policy is a default that both callers overwrite.** A third dispatch entry point that forgets `with_decode_policy` gets `Drop` for every `FromHeaders` contract, silently, because `Drop` is also the documented default. Nothing fails.

**The shape checks in `resolve_shape` are order-dependent.** `&[&[u8]]` is matched before the `raw` clause is read, which is what makes the retiring `batch(..)` clause compose with `raw` instead of being rejected. Reordering the branches for readability reverts that behaviour without any test noticing until the `subscriber_raw_batch` snapshot is regenerated.

**A hand-written definition silently gets no builder.** `Declared` is blanket-implemented for `IncludeDef` with `Settings = Self`, and the step traits exist only for `SubscriberBuilder`, so `.workers(..)` on a hand-written def fails on `Self::Settings: WorkersStep`. Correct, but the failure names an associated type the author never wrote. Inline detail.

**Broker extension traits must be written against the builder's `Src` parameter, not the definition's associated type.** The broker-authors guide shows the right shape (`SubscriberBuilder<Def, SubscribeOptions, (W, F, P)>`), which differs from the sketch in #234 (`Def: SubscriberDef<Source = ..>`). The guide is correct and the issue's sketch is not, because the source lives on the builder now. Nothing enforces that a broker crate reads the guide rather than the issue, and the wrong bound produces methods that never resolve.

### Places that resist mental simulation

**`Declared::Settings` is the seam that every mount now goes through, and both `include` methods changed shape around it.** `D::Form` still comes from the definition, but the form's `IncludeMount` / `RouterMount` bound is now on `D::Settings`, so the definition the mount machinery sees is never the one the user wrote. Every mount-site diagnostic reflects that, which is the visible half; the invisible half is that `HasSlots`, `BindSlots` and the six `*Def` forwards all had to be re-stated on the wrapper. They are, and `forward.rs` says why the associated types are spelled through `Def` rather than `Self` - that note is doing real work and should not be dropped.

**Six near-identical forwarding impls plus two macros.** `forward_common!` and `forward_outgoing!` keep it honest, but the file is a mechanical mirror of the definition traits, so adding a seventh form means remembering it exists. There is no compile-time link between the two sets.

### Knowingly approximated or left undone

- A raw batch takes no `Out` or `Seek` parameters; the macro rejects the combination with a message naming the alternatives. Stated in the PR body and in the guide.
- Double buffering stays legal (a negative bound would be needed), and the naming - `buffered`, not `max_size` - is what carries the meaning. That matches what #234 decided.
- Broker crates need an adoption wave for `Clone`, the `FromName` impls that make `#[subscriber(Kind)]` legal, and the extension traits over `SubscriberBuilder`. The conformance bound added here is what will surface the first of the three.

### Cross-check against #236

The two branches conflict only in `src/prelude.rs` and `src/runtime/mod.rs`, and only in the export lists. `crates/ruststream-macros/src/expand.rs` is touched by both and merges cleanly: #236 changes `outgoing_method` and `slot_bounds`, this one restructures the shape inference and the definition emission. No machinery is solved twice. Both PRs introduce a typestate over a builder - address segments there, settings slots here - and both carry the missing piece in the type so the error can name it; they share no code, and should not, since the two vocabularies have nothing in common. The one thing worth keeping in step is that both name the culprit rather than the machinery, and both do.

PR #276, opened 2026-08-29
Value-level definition constructors, and one include-only mount surface

### LINKED ISSUE #275: Unify the manual path behind one Handle trait and one subscriber constructor
Final design, third revision. Supersedes the constructor family and the body-trait set from the first two implementation rounds of this issue; everything lands inside the unreleased 0.7 line on the same PR.

### Problem

The first rounds shipped the manual path as a mirror of the form matrix: one constructor and one body trait per mount form (19 constructors, 10 body traits). The macro does not name these forms - the signature decides everything - and the manual path must reach the same economy: one public body contract, one mounting verb, with every axis (batch, raw, reply, injections, typed headers, seek, state) expressed by types the user already writes, not by vocabulary they must memorize.

### Proposed solution

**One body trait.** No associated types; every axis is a defaulted parameter:

```rust
pub trait Handle<In: ?Sized + Input, R = (), O = (), C = (), S = ()>: Send + Sync {
    fn handle(&self, input: &In, outs: &O, ctx: &mut Context<'_, C, S>)
        -> impl Future<Output: IntoVerdict<In, R>> + Send;
}
```

The return shape is computed from the axes and normalized by an `IntoSettle`-style conversion family. Canonical spellings:

```rust
Result<(), HandlerResult>                       // plain
Result<Reply, HandlerResult>                    // reply (R = Reply)
Result<(), Vec<HandlerResult>>                  // batch: Ok acks the page
Result<Vec<Reply>, Vec<HandlerResult>>          // batch reply
```

Settle-carrying spellings stay accepted exactly as today (`Result<(), Settle>`, per-element `Vec<Settle>`), so `and_after` continuations remain a return capability with the unchanged at-most-once contract. An illegal input-reply combination is unrepresentable: the return type is derived, not chosen, and the bounds carry `#[diagnostic::on_unimplemented]` guidance. Per-element vectors are exactly page-length; a mismatch is a documented panic (message names the subscription and both lengths) handled by the existing panic failure policy.

**Input axis.** `T` (decoded), `[T]` (batch), `[u8]` (raw), `[&[u8]]` (raw batch), `Message<Headers, Payload>` (typed headers), `[Message<Headers, Payload>]` (batch with headers). `Message` is a transparent pair type (public `headers` / `body`); when it is the input, the core decodes headers in the same stage and under the same `on_failure(decode = ..)` policy as the payload, and the body receives ready values. The pairing is structural: each consumer declares the contract it reads. `to_typed` becomes crate-private. The existing `Message` metadata trait and derive are renamed `MessageInfo` to free the name.

**Injections arena.** The second argument. Bodies without injections take `&()`. Slot bodies bound one opaque generic:

```rust
impl<O> Handle<Order, (), O> for Mirror
where
    O: Outs<(Primary, Shadow)>,
    O::Out<Primary>: Publish,               // capability bound is mandatory per marker
    O::Out<Shadow>: TransactionalPublisher, // no bound, no methods
{
    async fn handle(&self, order: &Order, outs: &O, ctx: &mut Context<'_>) -> Result<(), HandlerResult> {
        outs.get(Primary).message(&event).to("mirror").publish().await?;
        // ...
    }
}
```

The arena is fully typed and built statically by the include-site `.out(..)` chain (the router cons-list pattern); binding is order-free, and a missing, duplicate, or extra bind, or a policy weaker than the declared capability, fails to compile naming the marker. No type erasure anywhere; access is GAT-based and monomorphized. Handlers are constructed by the user with their own dependencies only; the framework delivers the arena per call.

**Seek.** Not a form and not an injection: position values live on the broker context axis (`ctx.position()`, `ctx.seek(..)` where `C` provides them). A source that cannot seek, or seeks with a different kind, fails at include. The `Seek(..)` parameter token remains as `macros`-feature sugar expanding to the same context access; the dedicated runtime mount forms are removed.

**State.** Untouched. One global application state travels through the framework exactly as today: the `S` axis on `Context`, `ctx.state()`, the `State` extractor and the `FromRef` derive all remain.

**Routing.** One constructor and a chain; the include site binds slots:

```rust
b.include(subscriber("orders", Audit).build())
 .include(subscriber("mirror", Mirror::new(deps)).build())
 .out(Primary, publish_policy)
 .out(Shadow, tx_policy)
 .build()
 .include(
     subscriber("orders", Confirm::new())
         .reply().to("confirmations")
         .publisher(reply_policy)
         .batch(nonzero!(8))
         .workers(nonzero!(12))
         .build(),
 );
```

`.reply()` alone takes the destination from the reply type's `OutgoingDestination`; `.reply().to(name)` names it; `.publisher(..)` attaches the policy on the definition chain (policies are connection-free). `.batch(n)` mounts native pages, `.buffered(n, window)` builds client-side pages; both feed the same `[T]` body. A chain step contradicting the body (`.reply()` on a non-reply body, `.batch(..)` on a non-slice body) fails to compile. `.build()` terminates both chains; the include-site binder replaces `.mount()`. `include` stays the single mount point.

**Documentation is on by default** under the `asyncapi` feature; the per-registration opt-out is `.undocumented()`. The `JsonSchema` obligation is checked at include only for definitions still in the documented typestate, and the error names both exits (derive `JsonSchema`, or chain `.undocumented()`). Deliberate exception, recorded here: enabling `asyncapi` can stop compiling a definition whose message type lacks `JsonSchema` and which does not opt out - the price of documented-by-default, accepted.

**Unchanged surfaces.** The publish builder (`.message(..)`, optional `.headers(..)`, dynamic destinations, key-wise `base_headers` merge) and the `PublishedThrough` dictionary; `Layer` / `DynLayer` middleware; `PublishTransform`; the broker contract; the state subsystem; the macro feature (not rewritten here, but the constructor API must not block a later macro emission onto it).

**Removed from the public vocabulary.** The 19 first-round constructors and their `_in` twins; the body-trait set (`Reply`, `BatchReply`, `SliceHandler`, `RawSliceHandler`, `SliceHandlerWithHeaders`, the `Slots*` family); `type Verdict` and the verdict gate; `BatchResult` in user returns; the runtime-side headers opt-ins (`documented_headers`, `reply_message`, `reply_headers` - subsumed by the axes); public `to_typed`. `Handler` remains as the middleware author surface; the definition traits remain as the macro target and SPI.

### Alternatives considered

- Type-erased slots (`Out<Marker>` over `dyn`): a vtable call and a boxed future per publish on the hot path; rejected.
- Slots through `Context` type parameters, or a type-map lookup on the context: signature noise, or a runtime miss for a misuse; rejected.
- Publishers as handler struct fields wired by user factories: injection must come from the framework, not through the user's constructor; rejected.
- An associated `Verdict` type, or the verdict as a trait parameter: replaced by the computed and normalized return, which also makes illegal combinations unrepresentable.
- A nominal `MessageHeaders` trait pairing each payload with one contract: replaced by the structural `Message<Headers, Payload>` axis (consumer-driven contracts).
- Moving `and_after` continuations to a context method: features must not be lost; the return capability stays.

### Additional context

- Breaking inside the unreleased 0.7 line; the broker 0.7 compatibility branches are also unmerged, so the fleet adopts the final shape in one pass.
- Zero cost holds: no `dyn`, no happy-path allocations (per-element vectors exist only on the failure path), the arena is built once at startup, everything monomorphizes.
- The completion artifact is the parity test rewritten over the new axes: every input kind, reply shape, injection set, seek, typed headers, and the settings chain, mounted on both surfaces; a missing spelling fails the build.
- The guides, the manual examples and tests, and the Cursor rules follow in the same change (see #274, #276).

powersemmi:
# Description

Implements the final design of #275 (body plus its two amendment comments): the manual path is one body trait and one mounting verb.

`Handle<In, R = (), O = (), C = (), S = ()>` carries every axis as a defaulted parameter, with the verdict computed from the axes rather than chosen: `Result<(), HandlerOutcome>` for a plain body, `Result<R, HandlerOutcome>` for a reply, and the page forms with per-element `Vec<HandlerOutcome>` vectors (exact page length, mismatch is a documented panic under the panic policy). `HandlerOutcome` is the public outcome type - the broker status plus an optional post-settle continuation (`and_after`), which runs strictly after the settle; the plain status enum is crate-private. The input axis spells batch, raw, and typed headers (`Message<Headers, Payload>`, decoded by the core under the decode failure policy); the injections arena is typed, built statically by the include-site `.out(..)` chain, with a mandatory capability bound per marker and direct naming of broker-defined live values; seeking rides the broker context axis; the state subsystem is untouched.

Mounting is `subscriber(source, body)` plus the settings chain (`.reply()[.to(..)]`, `.publisher(..)`, `.batch(..)`, `.buffered(..)`, `.workers(..)`, `.on_failure(..)`, `.start_at(..)`, `.build()`), with `include` the single mount point. Documentation is on by default under the `asyncapi` feature with a per-registration `.undocumented()` opt-out - the recorded deliberate exception to feature additivity. The publish builder, the `PublishedThrough` dictionary, `Layer` / `DynLayer`, `PublishTransform`, the broker contract, and the macro feature are unchanged; the definition traits remain the macro target.

The intermediate round-two surface (the per-form constructor family and body-trait set) is removed in the same line. The parity test mounts every axis combination on both surfaces, so coverage is checked by the compiler. The macro-free examples, the manual tests, the guides (en/ru/zh), and the Cursor rules all follow.

Fixes #275

## Type of change

The removal lands inside the unreleased 0.7 line, so no extra version bump is taken.

## Checklist

      `examples/` where applicable


powersemmi:
Round three supersedes the round-two surface on this branch: the 15-constructor family and the body-trait set are gone, and the manual path is one trait and one constructor, per the final design in #275 (body + the two amendment comments).

| Axis | Spelling |
|---|---|
| Body contract | `impl Handle<In, R = (), O = (), C = (), S = ()>`, no associated types |
| Verdict | computed: `Result<(), HandlerOutcome>`, `Result<R, HandlerOutcome>`, `Result<(), Vec<HandlerOutcome>>`, `Result<Vec<R>, Vec<HandlerOutcome>>`; `and_after` returns `HandlerOutcome`, the continuation runs strictly after the broker settle |
| Input | `T` / `[T]` / `[u8]` / `[&[u8]]` / `Message<Headers, Payload>` / `[Message<H, P>]`; headers decode with the payload under the decode failure policy |
| Injections | `outs: &O` with `O: Outs<(markers)>`; capability bound per marker is mandatory; a broker-defined live value is named concretely in the entry (`Outs<(Slot<Lanes, LaneRouter, Enc>,)>`) and reached through the slot's `Deref` |
| Seek | the broker context axis (`ctx.position()`, `ctx.seek(..)`); no dedicated mount forms |
| State | untouched: the `S` axis, `ctx.state()`, `State` / `FromRef` |
| Mounting | `subscriber(source, body)` with `.reply()[.to(..)] / .publisher(..) / .batch(..) / .buffered(..) / .workers(..) / .on_failure(..) / .start_at(..) / .build()`; include-site `.out(Marker, policy)..build()`; documentation on by default under `asyncapi`, opt-out `.undocumented()` |

Removed from the public vocabulary: the round-two constructors and their `_in` twins, `Reply` / `BatchReply` / `SliceHandler` / `RawSliceHandler` / `SliceHandlerWithHeaders` / the `Slots*` family, `type Verdict`, `.reply_raw()`, the runtime headers opt-ins, public `to_typed`, and `HandlerResult` itself (now crate-private; the public outcome type is `HandlerOutcome`). `Handler` stays as the middleware author surface; the definition traits stay as the macro target.

Deviations from the letter of the amendments, forced by stable Rust and recorded here:

- A GAT equality binding (`Out<Lanes> = LaneRouter`) is unreachable on stable (coherence rejects the marker-to-value lookup impls), so the concrete live type is named directly in the arena entry and reached through `Deref`; same semantics, different spelling.
- The `Good<R>` type function hits the same wall, so `R` is the whole Ok side: a batch reply body writes `Handle<[T], Vec<Reply>>`. The four canonical result shapes hold literally.
- Whether a reply wire is raw or encoded is selected by the publisher-step form (`Bare(policy)` vs the typed attachments), since a pure type function over `R` collides with `Vec<u8>: Serialize`.

The parity test (`src/runtime/handle/parity_tests.rs`) mounts every axis combination on both surfaces, so coverage is checked by the compiler. Gates on the head commit: `just ci` green (fmt, clippy `-D warnings` on all targets and features, unit + doc + integration suites, `--no-default-features`), UI snapshots 43/43, and the docs pipeline green locally (`lint_doc_fences`, `check_doc_i18n`, `check_doc_parity`, `properdocs build --strict`, en/ru/zh).

powersemmi:
Squash-merged into docs/cursor-rules-refresh as a44ccbe (with the macro rewrite following as ff94dd8); the branch is deleted.

PR #279, opened 2026-09-01
Rewrite the subscriber attribute onto the unified manual path

### LINKED ISSUE #275: Unify the manual path behind one Handle trait and one subscriber constructor

powersemmi:
# Description

Rewrites the `#[subscriber]` expansion onto the unified manual path, making the attribute sugar over the public value-level API. Fixes #275.

Every extractor, include-site chain (`.publisher(..)`, `.out(marker, ..)`, `.build()`, deferred `.name(..)`) and derive keeps compiling and behaving as before. The attribute's surface changes in the sanctioned ways listed below; everything else that changes is the emission:

- One `impl Handle<In, R, O, C, S>` around the user's body replaces the seven body-trait families (`Handler`, `SliceHandler`, `RawSliceHandler`, `InjectCall`, `PublishingCall`, `BatchInjectCall`, `BatchPublishingCall`). The user's return vocabulary is lowered into the fixed computed verdict through the existing `IntoOutcome` seam plus two new page-verdict helpers, with a uniform page outcome fanning out only on the failure path and its continuation riding the last element (at-most-once, unchanged).
- The generated `Declared` builds the same sealed definition the `subscriber(..) .. .build()` chain produces, so the runtime's own definition impls drive every mount; the macro implements none of the definition traits on its output any more, and the definition-trait emission (`expand/legacy.rs`) is deleted outright - the generation is unified, with no per-form remainder.
- A handler with `Out` parameters keeps the unit struct as the include-site value and instantiates the arena-carrying definition in `BindSlots::bind`, so the concrete publisher types are still inferred from the attached policies and the slot shell structs disappear.
- Schemas, `MessageInfo` metadata, the reply entry and slot dictionaries are captured by the existing autoref probes at the expansion site and ride the definition as data (a new hidden `Probed` documentation state), so the attribute keeps its schema-if-derivable behavior and never inherits the value path's documented-by-default `JsonSchema` obligation.
- The emission decides no lane and no form: the input axis is projected off the type (`<T as Input>::Axis`), the mount form off the axis and the reply type (`EagerForm` / `SlotForm` / `ReplyRoute`), so the macro's only shape question left is `&T` versus `&[T]`.

## The pipeline lanes are selected by the type

The raw path stops being a spelling of the signature and becomes a property of the types, under one dictionary: `Deserialize` / `Serialize` mean the framework's codec does the work, `Deserialized` / `Serialized` mean the user's own type already did it.

- **`Deserialized` (input).** A type constructs itself from the payload bytes (`fn from_payload(&[u8]) -> Result<Output<'_>, Error>`, a lifetime-carrying GAT so the view borrows the broker's buffer with zero copies). The core never decodes on this lane - no codec runs - and a failed construction settles under the subscriber's `on_failure(decode = ..)` policy, the same rung a codec decode failure lands on, so a validating format (flatbuffers verification, capnp reading, a framed protocol) fits today. `#[derive(Deserialized)]` covers the obvious newtype over `&'a [u8]`; the hand-written form is two short impls (`Deserialized` + `Input`), documented on the trait. The page spelling comes for free: one blanket `impl Input for [E]` projects the element's axis to its page axis, so `&[Frame<'_>]` mounts off the same two impls (a per-type slice impl would be an orphan-rule violation downstream anyway).
- **`Serialized` (outgoing).** A type carries its own wire bytes (`fn bytes(&self) -> &[u8]`), and the value leaves byte-for-byte - no codec, nothing re-encoded, no copy beyond handing the bytes to the broker - as a reply through a bare publisher, and through every typed publish entry: `message(&value)` reads the wire off the value's type via the `MessageWire` projection (a blanket over `Serialize` picks `EncodedWire`, per-type impls pick `SerializedWire`; no blanket over `Serialized`, so a type with both impls is a definition-site conflict, pinned as a UI case), so arena slots, the `Out(..)` parameter, `TypedPublisher`, transaction scopes and templated destinations all serve both wires from the one entry. `#[derive(Serialized)]` covers the newtype over an owned buffer; the hand-written form is `Serialized` + the wire spellings (`MessageWire` with `Wire = SerializedWire`, `ReplyShape` with `Wire = SerializedReply`), documented on the trait. Naming `.with_codec(..)` on a serialized value does not compile - the codec would be silently ignored.
- **`Payload<'_>` and the `&[u8]` / `&[Payload<'_>]` special cases are deleted.** The form rule is uniform - `&T` one message, `&[T]` a page - and a bare byte slice is a curated compile error naming `#[derive(Deserialized)]`. In-process examples and tests declare their own named view types.
- **One `.reply()`, one `publish("dest")`.** The reply's wire is derived from the reply type through disjoint machinery impls: `Serialize` routes onto the encoded wire, `Serialized` onto the raw one. `Bare(policy)`, `DefaultBareReply` and the `Vec<u8>`-means-raw rule are deleted; a `Serialized` reply attaches the publish policy itself (`.publisher(MemoryPublish)`) or takes the broker default, an encoded reply attaches a `TypedPublisher` as before, and `.publisher(..)` rejects the wrong attachment shape for the reply's wire at compile time. `Vec<u8>` is an ordinary `Serialize` reply now (it encodes); bytes that must leave untouched ride a `Serialized` newtype.
- The lanes compose on both diagonals: a `Deserialized` input with an encoded reply (the old rejection is gone) and a decoded input with a `Serialized` reply (the gateway shape) both mount. A `Serialized` page reply stays unsupported: page replies publish through the reply codec, and the wire-per-family projection says so in its diagnostic.
- Raw outgoing types are first-class dictionary members under the ordinary rules: `impl PublishedThrough<Marker> for Export` (or `#[derive(Outgoing)]`) puts a `Serialized` type in a slot's declared set, the slot's typed entry publishes it - `out.message(&export).publish()`, destination off the `#[outgoing]` declaration (mandatory `.to(..)` without one, templates as for any model), the dictionary, declared-set and headers gates unchanged - and the generated document lists it under its own name; `.raw(..)` stays for bytes that carry no message type at all. The metadata marks the self-carrying lanes schema-free-by-design on both sides (a `Deserialized` input through its kind, a `Serialized` outgoing entry through the probes), so neither the generator's missing-schema warnings nor `Spec::messages_without_schema()` flag them.
- The axis stays open by construction: flatbuffers and capnp fit `Deserialized` as it is, and a future self-decoded lane (a schema-carrying format like protobuf, where the type still wants the codec's failure policy but owns its decoding) slots in as one more axis marker without touching the form rule.

## Batch forms gain a subscription-scoped context

The `C` axis of the batch forms is no longer pinned to `()`. A page spans many deliveries, so what a page body may reach is subscription-scoped data - and the new `BuildBatchContext<Msg>` seam builds the broker's batch context once per page from its first delivery: no dyn, no per-message cost, one construction per dispatched page. Keeping the batch context a separate type from the per-delivery one is what rejects a page body asking for per-delivery fields at compile time.

The in-memory broker is the reference: `MemoryBatchContext` carries the subscription's seeker under the existing `SeekHandle` key. Where to seek rides the elements themselves - a `&[Message<H, T>]` page reads the target off each element's typed header contract - and the integration test drives the maintainer scenario end to end: the page reads a `resume_at` header, settles, repositions through `ctx.context(SeekHandle)`, and the next page opens at the named position. What happens to in-flight deliveries during a rewind is deliberately not pinned down here; the memory broker's drain-and-replay semantics stand as its own behavior, not as contract. The reply- and injection-carrying batch forms keep the unit context for now (their defs never had a context axis); extending them is mechanical when a scenario asks for it.

Runtime changes backing this:

- The state axis leaves the value-definition types: state is a quantification point, not data, so it now lives only on `Handle` impls and the adapters quantify over it. This is what lets a state-generic macro handler mount on an app with any state while its definition stays a nameable type.
- `Docs` carries the probe-captured overrides; the sealed definition impls report them ahead of the axis-computed values (`None` fields defer, so the value path is unaffected) and now also forward `MessageInfo` name / description.
- The arena `Slot` gains a defaulted `Body` position holding an `Out` parameter's declared message set, and the entry's typed publish checks membership (`ContainsMessage`) exactly as the retired injection wrapper did - the third `Out` position keeps its compile-time narrowing on the unified emission, and the probed docs report the narrowed list as the handler's send operations.
- The self-deserializing page rides its own dispatch adapter (`DeserializedBatch`): the constructed views borrow the deliveries' payloads (one page `Vec`, as before), a failed element is settled by the decode policy and never reaches the body, and the constructed rest arrives as one contiguous page.
- Hidden macro seams: `probed_def` / `probed_reply_def` constructors, `ProbedDocs`, `page_verdict` / `uniform_page`, and a public-but-hidden `DefaultReply` re-export (generated types name it; it now serves both wires, keyed by the mount token).

Seek is no longer special-cased anywhere. The body-side seek surface is the broker's context keys: the delivery's position and the subscription's reposition handle are ordinary broker context fields read by `ContextField` keys, like any other per-delivery data. The in-memory broker is the reference implementation - `MemoryContext` carries both, published under the `Position` / `SeekHandle` keys, with the seeker resolved once per subscription open and cloned into the context allocation-free. A macro handler reads them with the existing `Ctx<K>` extractor; a manual body names the broker's context type and reads `ctx.context(KEY)`. Removed with the special casing: the `SeekContext` / `SeekableMessage` pair, the `Seek` injection parameter and its extractor type (the word is simply not recognized any more, like the retired `batch` / `raw` clauses), and the `Seek` / `BatchSeek` mount forms. `start_at(..)` and the `Seekable` / `Seeker` / `Positioned` capability contract are untouched. Broker crates adopt by shipping their own context type plus keys (and now a batch context type over `BuildBatchContext`), on the memory model.

Attribute-surface changes, all inside the unreleased 0.7 line:

- The `batch(..)` source wrapper and the `raw` flag clause are retired. The subscription form is inferred from the payload parameter's type alone, matched syntactically over its tokens: `&T` is one message, `&[T]` a page of them - and the lane is the type's own business. A bare `&[u8]` (or `&[&[u8]]`) parameter is a curated compile error naming `#[derive(Deserialized)]`; a source that lived inside `batch(..)` moves to the ordinary source position.
- `publish_raw(..)` is retired with a curated error naming its replacement: `publish("dest")` plus a `#[derive(Serialized)]` reply type. One clause, both wires.
- Batch header contracts move onto the `Message` input axis. The old form - `&[T]` plus a `Headers<Vec<H>>` parameter - is gone, together with its dedicated runtime machinery; the canonical input is `&[Message<H, T>]`, which the unified emission maps to the pair page axis. The single-message `Headers<T>` extractor is untouched.
- `on_failure(decode = ..)` on a raw handler is no longer rejected: the construction can fail now, and the policy is exactly what settles it.
- A handler input may carry an elided lifetime (`&Frame<'_>`, `&[Frame<'_>]`): the emission introduces its own lifetime for the `Handle` impl and takes the `'static` representative for the lifetime-free projections.
- Seeking from inside a body is not a parameter but a broker context field, read with `Ctx<K>` (above). A `Message<H, P>` input now combines with a declared `Out` message set - the old rejection existed only because the set kept the legacy emission.

Deliberate behavior amendments, all inside the unreleased 0.7 line:

- A macro batch handler now settles through the canonical page verdict: a per-element vector of the wrong length is the documented panic (naming the subscription and both lengths) instead of the old retry-the-remainder leniency, and a batch reply's element count is checked the same way.
- Deferred-name reply and slot definitions still mount; the one spelling that changes is `.name(..)` on a reply definition followed by a post-include `.publisher(..)` (the named chain now commits with the broker's default policy at include). No test or example used it.
- With the state axis off the definition types, a manual body generic over the state can no longer leave the constructor's state parameter to inference; the six manual tests using that spelling now write the defaulted state axis (`impl Handle<Order> for ..`), with every assertion unchanged.
- `.name(..)` on an unnamed reply/slot attribute definition, `Headers<T>` extraction, `Ctx<K>` projection, `State` injection, worker lanes, failure policies, `start_at(..)` and the AsyncAPI output are covered by the existing tests and unchanged.

With the legacy emission gone, everything that was `pub` solely as its target leaves the public surface: the per-form definition and call traits (`SubscriberDef`, `PublishingDef`, `BatchPublishingDef`, `InjectDef`, `BatchInjectDef` and their `*Call` halves), the dispatch adapters (`InjectHandler`, `PublishingHandler`, `Batch*Handler`, `TypedBatch`, `DeserializedBatch`, `SliceHandler`), the input-kind vocabulary (`Decoded`, `Provided`, `DecodeWith`, `InputKind`; `RawBytes` and `RawSliceHandler` are deleted outright), `BatchResult` / `IntoBatchResult`, `FromStartup`, `EncodeReply` / `ReplySink`, and the typed injection wrapper (`TypedSlot`, deleted) are crate-internal now. The `Out` signature marker stays public as pure vocabulary (a phantom; nothing constructs it), and what the emission and the derives still name in user crates stays public: `Handle`, `Declared`, the settings builder, the sealed value types and probed seams, the forms tokens, the slot vocabulary (`OutSlot`, `OutMessages`, `PublishedThrough`, `ContainsMessage`, `HasSlots`, `BindSlots`, `SlotPublisher` - also the broker crates' capability-graft point), the probes, and the new lane vocabulary (`Deserialized`, `Serialized`, `MessageWire` with `EncodedWire` / `SerializedWire`, `ReplyShape`, `EncodedReply` / `SerializedReply`, the deserialized axis markers, `BuildBatchContext`).

The error surface was re-snapshotted (`RUN_UI_TESTS=1`): every `#[diagnostic::on_unimplemented]` guidance line survives verbatim; the declared-set cases point at the exact offending type in the signature. The clause-driven and seek cases are gone with their words; the retired `publish_raw` keyword pins a curated error naming the `Serialized` replacement, the bare-slice spellings pin the errors naming `#[derive(Deserialized)]`, and an unserializable encoded reply now fails at the include with the reply-route bound naming the type. A `message(..)` of a type on neither wire is pinned as the neither-wire case - the rejection names `MessageWire` next to serde's missing-derive guidance (the wire trait's own both-lanes note renders where rustc reports the trait obligation directly; the method-call path leads with serde's note, the right fix for the common case) - and a type deriving both `Serialize` and `Serialized` is pinned as a definition-site conflict of the wire impls.

## Type of change

## Checklist

      `examples/` where applicable


powersemmi:
Squash-merged into docs/cursor-rules-refresh as ff94dd8 (stacked on a44ccbe); the branch is deleted.

PR #283, opened 2026-09-03
Carry the manual path in the prelude without the macros feature

powersemmi:
# Description

A service without the `macros` feature writes by hand what the derives write, and those impls
name types the prelude did not carry: the `Input` spelling over `SoloDeserialized` for a
`Deserialized` type; `MessageWire` over `SerializedWire` and `SerializedReply` for a `Serialized`
one; the `OutgoingDestination` forms (`CallerName`, `FixedName`, `NameTemplate`) and the
`MessageHeaders` contracts (`NoHeaders`, `WithHeaders`) of an outgoing declaration; the slot
dictionary (`PublishedThrough`, `OutMessages` with the `OutgoingMessageMetadata` it reports);
and the `FromContext` binding behind an extractor parameter. Each was one more
`use ruststream::runtime::{..}` or `use ruststream::{..}` line, so the manual path could not be
written from the glob the docs present as the one import a service makes.

The prelude now carries those names, taken from where they live: the traits and wire markers
from `runtime`, the declaration vocabulary from the crate root. The crate root gains no
re-export, and a derive sharing its trait's name keeps coming in under the `macros` feature only,
so the trait namespace is imported exactly once. The rustdoc examples showing the hand-written
form import from the prelude, the manual tests and examples drop the imports the glob now covers,
and the `.cursor` rules describe the same surface.

A new test pins it: with `macros` off and no codec feature, it spells both lanes exactly as the
traits' rustdoc does, imports nothing from the crate but the glob and the in-memory broker, and
mounts them as a service. The reduced-feature CI leg runs it.

## Type of change

## Checklist

      `examples/` where applicable

### DIFF
--- a/.cursor/rules/advanced-usage.mdc
+++ b/.cursor/rules/advanced-usage.mdc
@@ -66,9 +66,11 @@ alwaysApply: false
   constructor, the body trait `Handle` with its input spellings `Message` and `Deserialized`, the
   reply traits `Serialized` / `ReplyShape`, the injections arena `Outs` / `Slot` / `Publish`, plus
   `FailurePolicies` / `FailurePolicy`, `Workers`, `Unnamed`, `Out`,
-  `DefaultSlot`, `RouterDef` and `nonzero!`). The lane machinery the impls name -
-  `Input`, `SoloDeserialized`, `SerializedReply` - stays an explicit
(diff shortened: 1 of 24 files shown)


powersemmi:

PR #284, opened 2026-09-03
Bound slots with the broker capability traits on both paths, and give the memory broker its prelude

powersemmi:
# Description

Two vocabularies were sharing names. A handler body stated its slot's capability in a family of
the runtime's own - `Publish`, and on this branch the twins `TransactionalPublish` /
`OwnedTransactionalPublish` / `RequestReplyPublish` - which said the same thing as the broker
capability traits under different names, and occupied exactly the plain names a mount site wants
for a broker's publish policies.

They are separated now, and each half gets the one vocabulary it needs.

**A body names broker capability traits.** It imports `ruststream::prelude::*` and nothing else -
it never knows which broker runs it - and bounds its injected slot with the capability it drives:
`Out<impl Publisher>`, `Out<impl TransactionalPublisher>`, `Out<impl OwnedTransactions>`,
`Out<impl RequestReply>`, and on the manual path the same trait on the entry's wired value,
`where W: TransactionalPublisher, E: Codec + Send + Sync`. The typed operations each capability
unlocks stay exactly as they were, now as inherent methods on the arena entry gated by those
bounds: `message(..)` under the include site's codec and the marker's dictionary, `begin()`
returning the typed `TransactionScope`, `transaction()` returning the typed `TypedTransaction`,
`request(..)` through the entry's own delegation. A transaction opened on an entry still admits
exactly what the entry's own publish admits, so it cannot publish what the generated document
never declared. The raw capability methods stay reachable on the entry too, with the slot's
test-capture attribution.

The include-site check keeps its force and lands on the broker trait, so the `on_unimplemented`
notes on `TransactionalPublisher`, `OwnedTransactions` and `RequestReply` are back to naming the
`Out` form; the UI snapshots pin that a policy without the capability still fails at the mount
naming the marker, the policy and the missing trait. The attribute no longer special-cases any
capability name: every bound it reads lands on the wired value.

The core prelude drops `Publish` and the three twins and gains `TransactionalPublisher`,
`OwnedTransactions` and `RequestReply`, so a broker-agnostic body can state its bound from the one
glob.

**A mount site names policies, under uniform names a broker prelude provides.** The in-memory
broker gets one, `ruststream::memory::prelude`, shaped the way every broker crate's should be:
`ruststream::prelude::*` first, then the broker's own surface a service names (`MemoryBroker`,
`MemorySource`, `MemoryError`, the context keys), then the policies aliased to `Publish`,
`TransactionalPublish` and `Request`, plus the manifest of capability traits this broker
implements on its live values. `TransactionalPublish` is the same policy as `Publish` here,
because the in-memory publisher carries both transaction kinds; `Partitioned` is deliberately
absent from the manifest, since in scope it makes `msg.partition_key()` ambiguous with the
defaulted `IncomingMessage` method. Every example, test and snippet that mounts on the in-memory
broker now globs that prelude and names `Publish` / `TransactionalPublish` / `Request` where it
named `MemoryPublish` / `MemoryRequest`; bodies keep the core glob, and a file holding both is
served by the broker glob alone.

Because the core prelude exports nothing under those three names, a broker prelude can claim them
without shadowing anything a body relies on, and swapping brokers swaps the glob rather than the
mount sites. `docs/broker-authors` states that as the contract, with the memory prelude as the
worked example.

Pins: `tests/memory_prelude.rs` mounts through the broker glob and asserts the separation holds
(`Publish` is the policy type, `Publisher` the core trait);
`tests/manual_slot_capabilities.rs` drives each capability's typed form through a manual body
against the in-memory broker (commit and abort of a slot scope with attribution, an owned
transaction's buffer, a correlated request round trip); the UI snapshots reject a slot bound by
each capability against a policy whose live publisher lacks it, on both the manual and the
attribute path.

## Type of change

- [x] New feature (a non-breaking change that adds functionality): `ruststream::memory::prelude`
      raising the MSRV - a minor version bump pre-1.0): the `Publish` trait family is gone from the
      public API and from the prelude, and `TransactionScope` / `TypedTransaction` take the opening
      surface as a type argument, so `Transactional::begin` and `TypedPublisher::transaction`
      return those instantiations

## Checklist

      `examples/` where applicable


powersemmi:

PR #286, opened 2026-09-03
Close the documentation gaps the 0.7 broker adaptation surfaced

powersemmi:
# Description

Documentation follow-ups reported by the broker crates while they were re-adapted to the 0.7 core.
Each item is a place where the docs, or code that doubles as a documentation snippet, no longer
matched the API.

- **Seeking.** The subscribers page ended its Seeking section with a sentence about a clause the
  page never introduced, and the compiled snippets showing that clause were orphaned. The section
  now separates opening a subscription at a chosen position from repositioning a live one from a
  handler, and the snippets are wired into the first. The same gap in the Cursor usage rules is
  closed.
- **`Out` slot extension.** A handler body holds the arena entry, which dereferences to the live
  value and never to the slot wrapper, so the extension the broker-authors guide described was
  unreachable from a body and a bound over it did not resolve. The graft moves to the entry in both
  snippet sources and in the rustdoc, and the guidance now separates a router-shaped capability
  (hands out a publisher, loses the slot's attribution to the broker's publish log) from a
  step-shaped one (rides the entry's typed publish path and carries its argument as a header).
- **The broker prelude.** The guide gains a section for it: the four layers in order
  (`pub use ruststream::prelude::*`, the crate surface the examples name, the publish policies
  aliased to the uniform mount-site names `Publish` / `TransactionalPublish` / `Request`, the
  capability manifest), the split between a handler body that imports the core prelude and bounds
  an injected slot with a broker capability trait (`Out<impl Publisher>` and friends) and a routes
  file that imports the broker's, and the probe that pins both halves - the bound still arriving as
  the core trait, the mount-site name still being the policy.
- **Borrowed context values.** A key over a position that is not `Copy` borrows through
  `Field::Value<'a>`; only the value behind the `Ctx` extractor has to be owned.
- **Page handlers under the harness.** Injecting through the harness drives each message to a
  settle, which closes the client-side buffer's page, so a test written that way never sees a page
  longer than one element. The testing guide and the adapter's own docs name the shape that does.
- **Stale rustdoc and template.** The `Ctx` doc comment named a return type the public API does not
  have, and the memory template named a router terminal that does not exist.
- **The worked NATS example** matched an older shape of the crate: a cell-backed broker, a unit
  `Closed`, two capabilities described as missing that are implemented, and a pinned client
  version. It now follows the crate's current shape and points at the broker's own docs for the
  client version.

Documentation pages are updated in all three languages.

Fixes # (no issue)

## Type of change

## Checklist

- [ ] I have added tests that validate my fix or new feature
      `examples/` where applicable

The two changed test files are the compiled sources of documentation snippets; they keep their
assertions, including the slot attribution boundary. No new behaviour is introduced, so no test is
added.


powersemmi:

PR #287, opened 2026-09-03
Make the page size the one mandatory batch parameter, forwarded to the broker

powersemmi:
# Description

A page handler's mount site had two ways to size the page it consumes and no way to require
either. `batch(n)` was a runtime cap: the broker sized its pages however it liked, and the
framework sliced whatever came back into chunks of at most `n`, settling each on its own. Next to
it, `buffered(max, wait)` sized the same page from the other end, which is why the two had to
exclude each other. And a page handler mounted with neither ran on whatever the broker happened
to hand over.

Slicing puts the framework between the broker and the body for no gain. A page could reach a
handler in pieces its broker never sent; a replying page committed one transaction per chunk
instead of per delivered page; and the number a service wrote was a bound on a page size nobody
had chosen. The second spelling made it worse by naming an adapter - `Buffered` is broker-author
machinery, and a service that reaches for it is answering a question its broker should have
answered.

The page size becomes the subscription's own parameter, passed down instead of applied.
`BatchSubscriber::batches` takes it:

```rust
fn batches(&mut self, size: NonZeroUsize) -> impl Stream<Item = Result<Self::Batch, Self::Error>> + Send + '_;
```

The runtime hands the mount site's size straight to the broker, which translates it into whatever
its client already speaks (`XREADGROUP COUNT`, a pull batch, a poll limit), and the page a body
sees is exactly the page the broker delivered - never a slice of it. The chunking in the body
adapter, the reply-side chunking in the publishing dispatcher, and the definition-carried cap
behind them are gone.

Naming the size becomes mandatory, and type-carried. A registration whose page-size slot is still
open implements no `PageSized`, so a page handler mounted without `.batch(nonzero!(n))` fails at
`include` / `build` with a message naming the fix, next to the existing refusal of `.batch(..)` on
a body with no page. Both directions are pinned by trybuild.

```rust
#[subscriber]
async fn reconcile(orders: &[Order]) -> Vec<HandlerOutcome> { /* .. */ }

b.include(reconcile.name("orders").batch(nonzero!(6)));
// on a Redis broker the size stays the core's word and `.block(..)` is the broker's, chained
// after it on the subscription source:
// b.include(reconcile.name("orders").batch(nonzero!(6)).block(Duration::from_secs(5)));
```

`.buffered(..)` leaves the mount site with the setting it duplicated. `Buffered` and
`BufferedSubscriber` stay public and grow the entry point a broker crate needs
(`BufferedSubscriber::new`): a transport with no native pages implements `BatchSubscriber` by
delegating to the adapter inside the broker crate, with the deadline as its own constructor
argument and the size arriving per subscription. Every broker offers pages either way, and the
mount site cannot tell which. `Seekable for BufferedSubscriber` stays, so a delegating broker
keeps its seek capability through the wrapper.

The in-memory broker honours the size natively and `set_batch_limit` goes away. The conformance
batch suite now opens its subscription at a size smaller than the run and fails a broker whose
pages come back longer.

`assert_page_sizes` follows the meaning: it reports the pages the body was handed across the run,
which is where the broker's answer to `batch(n)` is visible - a log of three replayed under
`batch(2)` reads `[2, 1]`, two calls, because the broker built two pages.

Two smaller things ride along in the branch's earlier commits. `BufferedSubscriber` gained
`Seekable`, so a page subscription on a broker that pages through the adapter can still open at a
chosen position. Two quiescence gaps closed that made a page assertion racy: a page's record is
written before the settlements that release the in-flight count, and a reposition is held in
flight from the moment the seek resolves. The `#[subscriber(EXPR)]` source error also says why a
free-function source is rejected now, which its two siblings already did.

Fixes # (no issue: reported from two broker crates' 0.7 adoption)

## Type of change

Breaking within the unreleased 0.7 line, and every broker crate adapts on its open branch:
`BatchSubscriber::batches` takes the page size, a broker drops any size knob of its own, a
transport with no native pages delegates to `Buffered`, and `.buffered(..)` call sites become
`.batch(..)`.

## Checklist

      `examples/` where applicable

Also run: both trybuild suites with `RUN_UI_TESTS=1` (snapshots regenerated), the three
documentation scripts, and `properdocs build --strict`.


powersemmi:

PR #288, opened 2026-09-03
Let the batch forms carry the broker's page context

powersemmi:
# Description

A page body that both answers through the reply clause and names the broker's subscription-scoped
context did not compile, while the same body without a reply did:

```rust
#[subscriber("orders", publish("digests"))]
async fn digest(page: &[Order], ctx: &mut Context<'_, MemoryBatchContext>) -> Vec<Digest> { .. }
```

The batch definition traits behind the reply and the injections forms carried no context axis at
all and spelled the unit type into their call signature, so naming a broker page context closed the
`publish(..)` clause and vice versa. The single-message definitions have carried that axis since
the typed-context work, so this was a gap in the batch half only, and it hit the manual path
identically (`Handle<[Order], Digest, (), MemoryBatchContext>` mounted through
`subscriber(..).reply().to(..).build()`).

The batch definitions now carry the context axis their single-message counterparts already have,
and the mounts thread it from the definition to the dispatch loop, so the runtime builds the page
context off the page's first delivery exactly as it does for a settling page. The reply publisher
was already generic over the context it reads while publishing, so nothing new is checked at run
time: the combination is either well-typed or it is a compile error naming the missing bound.

The forms that were pinned the same way are fixed together, since they are one axis and not four:
the plain page reply, the pair page reply, a reply with `Out` slots, an `Out` arena without a
reply, and the client-side buffered source under any of them.

## Type of change

The definition traits are crate-internal (their modules are private and nothing re-exports them),
so the added associated type is not a public-API change; the change lands inside the unreleased
0.7 line either way.

## Checklist

      `examples/` where applicable

## Tests

- The parity suite gains the missing cells, so a lost combination fails the build again: a page
  reply reading the page context, the pair-page form of it, the three-axis cell (reply, arena and
  context together), the arena-without-reply cell, and the buffered source under a page reply.
- The matrix suite drives one of them against a context the runtime built off a real delivery, so
  the handle the context carries is exercised rather than only typed.
- A new integration suite runs both shapes end to end on the in-memory broker through the test
  harness: a replying page reading the page context, and one repositioning the subscription through
  it before answering.

The docs already described the page context as available to any page body, so the fix makes the
existing text true rather than needing new text.

### DIFF
--- a/src/runtime/app/include/forms_batch.rs
+++ b/src/runtime/app/include/forms_batch.rs
@@ -5,7 +5,10 @@ use serde::Serialize;
 // The typed default-reply commits build a `TypedPublisher`, so the codec import is gated.
 #[cfg(any(feature = "json", feature = "cbor", feature = "msgpack"))]
 use crate::codec::DefaultCodec;
-use crate::{BatchSubscriber, Broker, Connected, PublishPolicy, Subscriber, SubscriptionSource};
+use crate::{
+    BatchSubscriber, Broker, BuildBatchContext, Connected, PublishPolicy, Subscriber,
+    SubscriptionSource,
+};
 #[cfg(any(feature = "json", feature = "cbor", feature = "msgpack"))]
 use crate::{DefaultPublish, Publisher};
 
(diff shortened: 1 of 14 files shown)


powersemmi:

PR #289, opened 2026-09-03
Ride the TestApp harness in the core's own integration suites

powersemmi:
# Description

The crate ships a testing harness for services (`TestApp`), and most of its own integration suites
predated it: they collected what a handler saw in `static` counters, polled the count with a
`wait_for` loop over a started app, and tore it down by hand. That style pins "something eventually
happened" rather than what arrived, how it settled, and how many times - and it is the style the
core was teaching by example.

This moves every suite whose subject is the application surface onto the harness: handler bodies
receiving values, extractors, settlement outcomes, retry and delayed redelivery, failure policies,
batch pages, workers and keyed partitioning, replies and publish middleware, scope codecs,
post-settle continuations, and a seek issued from a page body. The injections settle the whole
reaction before they return, so the migrated assertions are exact counts, exact values and exact
outcome sequences instead of "at least one".

Suites whose subject IS the lower level keep it and now say so in a one-line comment: the started
app and its `RunningApp` (start, readiness, teardown, the health probe), `run_until` and the
shutdown drain and its deadline, the lifespan ladder, the CLI, the exporter tests that need a
health probe or a bus killed under a handler that is still holding its delivery, and the
manual-path parity suites whose gate deliberately excludes the `testing` feature.

The harness gained what those migrations needed, plus the fixes for three ordering gaps its wider
use exposed (a batch page recorded after the settle that releases the quiescence wait, a subscriber
built before the app reached the harness never balancing the in-flight count, and a reposition's
replay counted only once the stream applied it).

Fixes # (no issue)

## Type of change

## Checklist

      `examples/` where applicable


powersemmi:

PR #290, opened 2026-09-03
Build the reply wiring on the mount site's chain, and drop TypedPublisher from the public surface

powersemmi:
Stacked on #284; review that one first, and merge it before this.

# Description

`TypedPublisher` was a value a service constructed itself and handed to `.publisher(..)`. Everything
it carried is wiring the mount site already decides - the codec the reply encodes with, the static
`PublishTransform` stack that runs on it, and, through `.transactional()`, the broker transaction a
page's replies ride - so all of it moves onto the chain the mount site already writes, in the same
vocabulary as the subscriber settings. A service now names only the policy its broker prelude
exports:

```rust
// attribute path, include site
b.include(settle).publisher(TransactionalPublish).codec(JsonCodec).transactional();
b.include(confirm).publisher(Publish).transform(StampSource);

// manual path
subscriber(source, body).reply().to("confirmations").publisher(Publish).codec(JsonCodec).build();

// router
Router::new().include(confirm).publisher(Publish).transform(StampSource).build()
```

`b.include(confirm).publisher(Publish)` alone keeps the default codec, and a `publish("name")`
handler with no `.publisher(..)` keeps using the broker's `DefaultPublish` policy as before.

**The steps are type-carried, like the subscriber settings.** Each fills its own slot in the
wiring's type, so a second `.codec(..)` or `.transactional()` does not compile, and a byte-for-byte
(`Serialized`) reply carries the policy alone - none of the steps exist on that wire, because its
bytes leave unencoded and untransformed. Each gate is split in two so the error can say which of
those two things went wrong: the step trait says whether the wiring takes that step at all, and a
marker on the slot it fills (`CodecSlotOpen`, `PublishingDirectly`) says whether the slot is still
open. The gate is then an associated-type bound, which surfaces as a real trait obligation rather
than a method-probe rejection - so the `#[diagnostic::on_unimplemented]` guidance renders at the
offending call:

```
error[E0277]: this reply's codec is already named
29 |             .codec(CborCodec);
   |              ----- ^^^^^^^^^ `.codec(..)` fills the reply's codec slot, and it is filled
   = note: a reply encodes with exactly one codec: drop one of the `.codec(..)` calls
```

`.transactional()` on a policy whose live publisher is not a `TransactionalPublisher` fails at the
mount that named it, with the capability trait's own note. Three trybuild snapshots pin the three.

**The chain has no order to remember.** Naming the publisher closes no other step: the value steps
(`describe`, `undocumented`, `batch`) and the declarative settings (`name`, `workers`, `on_failure`,
`start_at`, `buffered`, `map_source`) all reach through the reply wiring, each keeping its own
"set once" typestate, so the two spellings below build the same definition. `parity_tests.rs`
mounts both of every pair.

```rust
subscriber(src, body).reply().to("out").batch(nonzero!(8)).publisher(Publish).codec(JsonCodec).build()
subscriber(src, body).reply().to("out").publisher(Publish).codec(JsonCodec).batch(nonzero!(8)).build()
```

**What the removed types were.** `TypedPublisher` and `Transactional` are what a paired wiring
produces at run time - the live reply sinks the dispatch publishes through - so they stay as
`#[doc(hidden)]` machinery and leave the documented surface and `ruststream::prelude`. Their
`PublishPolicy` impls move to the wiring value the chain builds; the constructors, `transform`,
`batch_transform` and `transactional` are gone.

**The typed transactions they carried move onto `PublishExt`.** Any publisher opens them:
`publisher.begin()` for the borrowed kind (the `TransactionScope`) and
`publisher.owned_transaction()` for the owned one (the `TypedTransaction`). The second name spells
out *owned* because a broker's own `OwnedTransactions::transaction` is in scope wherever this is,
and two same-named trait methods on one publisher would be ambiguous at the call. Both take the
publisher's own codec position, so a hand-driven transaction encodes with the default codec unless
the publish names one with `.with_codec(..)`; a slot's `begin()` / `transaction()` are unchanged and
still encode with the slot's codec and keep its `Admits` dictionary gate.

**The router's `.publisher(..)` stops being a terminal.** The chain continues into the wiring steps,
and `.build()` is the single commit - the same shape the slot-carrying router builders already had,
so `.publisher(..)` never commits anywhere now.

`.out(marker, policy)` is untouched: a slot entry takes the include site's codec and has no
transform stack or transaction of its own to name, and it never did - a `TypedPublisher` passed
there would not have satisfied `Out<impl Publisher>`. A message that needs a transform still gets it
from the reply path or from the app-wide `publish_layer`.

No behaviour change on the wire: the reply encodes with the codec the chain names and publishes
through the same builder.

## Type of change

      raising the MSRV - a minor version bump pre-1.0): `TypedPublisher` and `Transactional` leave
      the public surface and the prelude, `TransactionScope` / `TypedTransaction` take a codec
      position instead of a codec, and a router registration commits on `.build()` rather than on
      `.publisher(..)`

## Checklist

      `examples/` where applicable


powersemmi:

PR #291, opened 2026-09-03
Put every slot publish on the mount site's publish path

powersemmi:
Stacked on #290; review that one first, and merge it before this.

# Description

A publish through an injected `Out` slot went straight from the body to the broker. The app-wide
`publish_layer` chain never saw it - it was wired into the reply forms only - and a slot had
nowhere to name a transform of its own, so an outbox envelope or a publish-metrics layer simply
did not reach what a handler fanned out. The guide said otherwise ("an injected `Out` publisher
... stamped by the app-wide `publish_layer`"), which is what made the hole hard to notice.

Both halves are now true, and they are the same mechanism.

**A slot entry carries a publish pipeline next to its codec.** It is the app's own chain with that
slot's `.transform(..)` steps composed on top, resolved at the mount and paired into the arena at
startup - pair-time only, nothing per message beyond the layer itself. It sits above the
attributed leaf, so the harness's per-slot capture and the broker's publish log see the same
stamped message. A slot that names no transform in an app that adds no middleware keeps
`PublishIdentity` there, which is the bare leaf call the entry always made: same code, same error
type, no message rebuilt.

```rust
b.include(mirror).out(Audit, Publish).transform(OutboxEnvelope).out(Journal, Publish).build();
```

`.transform(..)` applies to the step the chain named right before it - a slot after `.out(..)`,
the reply after `.publisher(..)` on the mounts that carry both - so one registration writes both
sides and neither transform reaches the other's message:

```rust
b.include(confirm)
    .publisher(Publish).transform(StampSource)
    .out(Audit, Publish).transform(OutboxEnvelope)
    .build();
```

Before either call the step does not exist, and the gate is an associated-type bound rather than a
method-probe rejection, so the guidance renders at the offending call:

```
error[E0277]: this `.transform(..)` has no step to apply to
31 |             .transform(Envelope)
   |              --------- ^^^^^^^^ no `.out(marker, policy)` or `.publisher(policy)` call precedes it
   = note: a transform rides the step named right before it: `.out(Marker, Policy).transform(..)`
           for a slot, `.publisher(Policy).transform(..)` for a reply
```

**The transform is a trait of its own.** `OutTransform` is `apply(&mut Outgoing<'_>)` with no
`PublishContext`: a slot publish is issued by the handler body, not by the dispatch, so there is
no originating delivery for the entry to hand it - the body reads its own `Context` and puts what
it wants on the message. A type that belongs on both a reply and a slot implements both traits;
the two bodies are the same line.

**What stays outside the pipeline**, both of them publishes a body drives itself: a transaction
opened on a slot (`begin()`, `transaction()`) sends into the broker's transaction, and a
`request(..)` round trip waits for an answer instead of ending in a send. The entry reports those
in its own error type, so a body sees one error type either way.

**The type-carried piece is the entry's fifth parameter**, `Slot<Marker, W, Enc, Pipe, Body>`,
defaulted to `PublishIdentity`. A body's `Out<impl Publisher>` bound is unaffected - the attribute
path leaves the parameter generic for you - and a hand-written `Handle` body leaves it generic too
(`Pipe: OutPipeline`) to mount under any transform or app-wide layer; spelled without it, the entry
pins the plain path, which is what a mount with neither gives it.

**A router's slots take their own transforms and not the app-wide chain.** Binding a slot
instantiates the definition, and the entry's pipeline is part of that instantiated type, while the
app a router is mounted into (and the middleware it carries) exists only afterwards - a routes
function returning `impl RouterDef<B>` has nothing to resolve it against. `router/form_out.rs`, the
guide and the cursor rules all say so, and a handler whose slot publishes have to travel the
app-wide chain mounts on the broker scope with `b.include(..)`. Closing that would need either the
pipeline on `Router`'s own type (it would leak into every routes signature) or an erased pipeline
handle on the entry (a branch and a boxed future per slot publish); both are worse than the
boundary, and neither belongs in this change.

Two smaller pieces came with it: `TestApp::start` / `with_state` no longer demand an app without
publish middleware (the pipeline is already folded into the handlers it registered, so the harness
drives either), and `PublishedAssertions::with_header` reads what a transform or a layer stamped
on the way out.

Pins: `tests/out_transforms.rs` drives the layer over a slot publish (asserted on
`tb.out::<Marker>()` and on the broker's log), a transform on one of two slots, and a handler that
replies and publishes through a slot under one app-wide layer; the parity suite mounts
`.out(..).transform(..)` on both surfaces, with two slots and the transform on one of them, and a
registration transforming its reply and its slot in one chain; `tests/ui/out_transform_without_slot.rs`
pins the error above.

## Type of change

- [x] New feature (a non-breaking change that adds functionality): `OutTransform`, the
      `.out(..).transform(..)` step, the app-wide pipeline over slot publishes,
      `PublishedAssertions::with_header`
      raising the MSRV - a minor version bump pre-1.0): `Slot` takes the publish path as a fifth
      parameter (defaulted), so a hand-written body that mounts under a transform or an app-wide
      `publish_layer` names it; a slot publish through a pipeline reports a
      `PipelinePublishError`

## Checklist

      `examples/` where applicable


powersemmi:

PR #301, opened 2026-09-08
Let the routing suite accept a transport that cannot acknowledge

powersemmi:
# Description

`conformance::capabilities` accepts `AckError::Unsupported` wherever it settles a delivery
(`Ok(()) | Err(AckError::Unsupported) => {}`, eleven sites), and so does `harness::lifecycle`.
`harness::run_suite` - the routing contract every in-process stand-in runs - demanded a successful
`ack` at five sites and a successful `nack` at two. The two halves of one contract disagreed.

The cost lands on transports that settle nothing: ZeroMQ acknowledges nothing at all, MQTT has no
negative acknowledgement and no acknowledgement at `QoS 0`, Redis pub/sub and simple lists have
none, a file or stdio transport has nobody to give one to, Core NATS has none outside JetStream.
Their production `IncomingMessage::ack` correctly reports `AckError::Unsupported` - and their
stand-in could not, because the routing suite refused it. So the stand-ins reported success
instead, which is a stand-in lying about its transport: a handler returning `HandlerOutcome::retry()`
is exercised in process and silently loses the message in production.

The suite now reads settlement as what it is - something a transport either supports or does not -
at the sites that settle a delivery, exactly as the capability suites already read it. No new
public item is needed: the delivery's own answer to `ack` / `nack` is the probe, read where the
scenario depends on it.

Per site, and why:

- `ordering`, `publish_after_subscribe`, `headers_propagate` - the ack is hygiene; the assertion is
  order, late-subscription and headers. Tolerated.
- `ack_consumes_delivery` - tolerated, assertion kept: a transport with no acknowledgement consumes
  the delivery by delivering it, so "one publish, one delivery" is the same contract either way and
  still catches a stand-in that duplicates.
- `nack_without_requeue_drops` - tolerated, assertion kept: dropping is what such a transport does
  with every delivery anyway, so "it must not come back" holds for both answers and stays checked
  for both.
- `nack_with_requeue_redelivers` - **skipped**, not tolerated. This is the one scenario whose
  assertion IS the settlement. Accepting `Unsupported` and then reading a redelivery would assert
  nothing at best, and at worst pass a broker whose retries lose messages. A transport that reports
  the requeue unsupported has no redelivery to observe, so the scenario ends there. For a broker
  that does settle, the assertion is unchanged.
- An `ack` or `nack` that fails for any other reason still fails the run.

`src/testing/` was audited too. Nothing there requires a successful settlement: `dispatch` records
`settle.as_ref().map(HandlerOutcome::outcome)` - the handler's decision - **before** `settle_outcome`
touches the message, and an ack failure is logged, never propagated. `TestApp`'s `settled(..)` and
`assert_outcome(..)` therefore already pass for a transport that refuses settlement, and the
delivery is released to `Coordinator::consumed` from its `Drop`, which a refused settlement reaches
like any other. Only the doc needed correcting: `settled` said it "compares how the broker settled",
which is what a reader would take as the transport's answer.

Unblocks the honest form in `ruststream-zeromq` (both operations), `ruststream-rumqttc` (`nack`
always, `ack` at `QoS 0`), `ruststream-sea-file` (both, on both transports) and `ruststream-fred`
(pub/sub and simple lists).

## Type of change

Loosening only: no public signature changes, no new public items, no new required trait method. A
broker that passed the suite before passes it now.

## Tests

`tests/conformance_settlement.rs` drives `run_suite` against three stand-ins built from the same
routing (the in-memory broker's), differing only in what their deliveries claim:

- refusing every settlement with `Unsupported` (the honest form of a transport that settles
  nothing) - passes the suite; on `main` it fails at `harness.rs:203`, the first scenario, with
  `ack failed: Unsupported`;
- claiming every settlement worked while performing none (the lie the suite must still catch) -
  still fails, at the redelivery its claim promised;
- failing settlement with `AckError::Timeout` - still fails, since what is accepted is the one
  answer that names a transport without acknowledgement, not every error.

## Checklist

- [x] I have made the necessary changes to the documentation (rustdoc and the docs site, en/ru/zh)
      `examples/` where applicable


powersemmi:

PR #302, opened 2026-09-08
Keep the deferred retry when the transport cannot settle

powersemmi:
# Description

The broker-agnostic `retry_after` fallback settled the original delivery with `nack(false)` and
propagated the result with `?`, before spawning the deferred re-publish. On a transport whose
settlement is unsupported that `nack` returns `AckError::Unsupported`, so the `?` returned out of
the fallback and the deferred publish was never spawned. The delivery was neither retried nor
redelivered: the runtime promised a re-publish and dropped the message instead.

The shape covers every transport with no settlement of its own: MQTT at QoS 0, ZeroMQ, the file and
stdio transports, Redis pub/sub and simple lists. It was measured on MQTT against a real broker,
where the deferred copy appears at QoS 1 and never at QoS 0.

An unsettleable delivery is exactly the case where the re-publish is the only form the retry can
take, so the fallback now tolerates `Unsupported` from the drop and goes on to schedule the copy. A
settle the broker actively rejects (`Broker`, `Timeout`) keeps aborting: the delivery is still the
broker's to redeliver on its own timers, and a deferred copy on top of that would duplicate the
message rather than save it. Both cases are covered by a unit test; the first one hangs on `main`
waiting for a copy that never comes.

No signature changed: `settle_nack_after` is private, and the public surface is untouched.

## Type of change

## Checklist

      `examples/` where applicable


powersemmi:

PR #303, opened 2026-09-08
Route a batch retry_after through the delayed-redelivery fallback

powersemmi:
# Description

The batch settle path applied a `NackAfter` by calling `nack_after` on the delivery, while the
single-message path first checks `supports_nack_after` and otherwise runs the broker-agnostic
fallback (drop the original, re-publish a copy to its own source after the delay).
`supports_nack_after` defaults to `false`, which is what nearly every broker ships, so the batch
call returned `AckError::Unsupported`, the runtime logged `ack / nack failed`, and the retry was
gone: no deferred copy, no requeue, nothing. A `HandlerOutcome::retry_after` from a batch handler
was silently dropped on almost every broker.

The two settles were otherwise the same function, and that is what let them drift: a second place
to answer a settlement is free to answer it differently, and the delay fallback is exactly the part
that is easy to leave out. There is one place now. The batch path threads the scope's `Delivery`
(which carries the retry publisher) where it threaded only its task tracker, and calls the
dispatcher's `settle_outcome` at all five of its settle sites: the uniform and per-element arms of
`settle_batch`, both arms of `settle_split_batch`, and the decode rejection in `decode_batch`.
`Context::delivery` replaces `Context::tasks` and hands the delivery out for the scope's lifetime
rather than the context's, so the batch path can take it before lending `&mut Context` to the
handler without holding a borrow across an await.

Four tests cover the five sites; all four fail on `main`, each with the original never dropped and
no deferred copy ever published:

| test | site |
|---|---|
| `a_uniform_batch_retry_after_defers_a_republish` | `settle_batch`, uniform arm |
| `a_per_element_batch_retry_after_defers_only_its_own_element` | `settle_batch`, per-element arm |
| `a_split_batch_defers_the_rejected_and_the_accepted_alike` | `settle_split_batch`, rejected index and accepted remainder |
| `a_deferred_decode_rejection_is_republished` | `decode_batch`, decode rejection under a `retry_after` decode policy |

Three doc comments (`HandlerResult::NackAfter`, `HandlerOutcome::retry_after`,
`FailurePolicy::RetryAfter`) still said a broker without native delayed redelivery falls back to an
immediate requeue. That stopped being true when the deferred re-publish landed: an immediate
requeue is only what is left when no `retry_via` publisher is wired. Second commit, doc only.

No signature outside the crate changed: `settle_batch`, `settle_split_batch`, `decode_batch`,
`settle_outcome` and `Context` are all `pub(crate)`.

### Relation to #302

Same bug one level up, and the fixes compose rather than overlap. #302 makes the shared fallback
tolerate `Unsupported` from the drop and still schedule the deferred copy; this PR makes the batch
path reach that fallback at all. Routing both paths through one function is what makes #302's
decisions apply to batches for free, so this branch does not restate them.

The two touch `src/runtime/dispatch.rs` in different places: #302 rewrites `settle_nack_after`'s
body and doc, this PR only adds `pub(crate)` and three doc lines to `settle_outcome` above it.
Either order merges; if git does flag the adjacency, the later branch keeps both hunks.

## Type of change

## Checklist

      `examples/` where applicable

### DIFF
--- a/src/runtime/batch.rs
+++ b/src/runtime/batch.rs
@@ -20,13 +20,12 @@ use opentelemetry::{KeyValue, global};
 // reach decoding through `DecodeWith`.
 #[cfg(test)]
 use serde::de::DeserializeOwned;
-use tokio_util::task::TaskTracker;
 use tracing::{error, warn};
 
 use crate::IncomingMessage;
 
 use super::context::Context;
-use super::dispatch::Workers;
+use super::dispatch::{Delivery, Workers, settle_outcome};
 use super::failure::{FailurePolicies, FailurePolicy};
 use super::handle::Deserialized;
 use super::handler::{HandlerOutcome, HandlerResult};
@@ -375,9 +374,9 @@ where
         if accepted.is_empty() {
             return;
         }
-        let tasks = ctx.tasks().clone();
+        let delivery = ctx.delivery();
         let result = self.inner.handle_slice(&values, ctx).await;
-        settle_batch(accepted, result, &subscription, &tasks).await;
+        settle_batch(accepted, result, &subscription, delivery).await;
     }
 }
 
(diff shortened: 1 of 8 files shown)


powersemmi:

PR #304, opened 2026-09-08
Correct the settlement defaults the broker-authors page teaches

powersemmi:
# Description

Two errors on the page a broker author reads before writing anything, plus one omission of the same
kind found while reading the section around them.

**`nack_after`'s default.** The `IncomingMessage` sketch annotated it as `// Defaulted: a plain
nack(true)`. It defaults to `AckError::Unsupported` (`src/message.rs:281`), and deliberately so: a
transport that cannot hold a message back reports that rather than silently turning a back-off into
an immediate redelivery. An author reading the page wrote a broker they believed requeued after the
delay. That is the misunderstanding the recent settlement work has been unwinding, and this is
where it was being taught.

**`supports_nack_after` was missing from the sketch.** The runtime reads it before it will call
`nack_after` at all (`src/runtime/dispatch.rs`), so a broker that overrides `nack_after` alone
leaves the gate at its `false` default and the override is never reached: native delayed redelivery
that quietly does not run. The sketch annotates the other two defaulted methods inline for teaching
and simply did not mention this one, so the page gave no way to find out. Added, annotated as the
gate it is.

**The stale fallback claim.** "A broker that overrides neither defaulted method still works with
every runtime feature: `retry_after` falls back to an immediate requeue" - the same sentence
corrected in three rustdoc spots in #303. The runtime drops the delivery and re-publishes a copy to
the same source after the delay, through the `BrokerScope::retry_via` publisher, with the
retry-count header incremented; an immediate requeue is only what is left when no such publisher is
wired.

## What else was checked

The whole page was read against the code, not only the two lines reported. Everything else holds:
the `Broker` / `ConnectedBroker` ladder, `Subscribe`, `Subscriber`, `Publisher` (including
`base_headers` defaulting to `None`), `PublishPolicy::pair`, `DefaultPublish`, `partition_key`
defaulting to `None`, the `BufferedSubscriber` 10 ms default, and the per-delivery and batch context
contracts all match the source. The settlement defaults above were the only drift.

The `TestableBroker` paragraph further down carries a fourth claim of this kind ("treats ack/nack as
effectively a no-op"), which #301 is already rewriting. Left untouched here so the two do not
collide.

## Note on overlap

#301 does touch `docs/broker-authors/index.md`, contrary to what I was told when this was scoped -
but at line ~613, in the `TestableBroker` section. This PR changes lines 93-120. Roughly 500 lines
apart, no conflict in either merge order. #302 and #303 do not touch the page.

Code comments inside the fence stay English in all three files, as everywhere else in the docs; the
prose around them is translated.

## Type of change

## Checklist

      `examples/` where applicable

### DIFF
--- a/docs/broker-authors/index.md
+++ b/docs/broker-authors/index.md
@@ -98,7 +98,11 @@ pub trait IncomingMessage: Send + Sync {
     async fn ack(self) -> Result<(), AckError>;
     async fn nack(self, requeue: bool) -> Result<(), AckError>;
 
-    // Defaulted: a plain nack(true). Override when the transport has native
+    // Defaulted: false. The runtime reads this first and never calls
+    // nack_after without it, so override the pair together.
+    fn supports_nack_after(&self) -> bool;
+
+    // Defaulted: AckError::Unsupported. Override when the transport has native
     // delayed redelivery (JetStream NAK with delay); handlers reach it through
     // HandlerOutcome::retry_after.
     async fn nack_after(self, delay: Duration) -> Result<(), AckError>;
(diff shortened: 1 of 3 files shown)


powersemmi:

PR #305, opened 2026-09-08
Cut the broker-authors page's sketches from code that compiles

powersemmi:
# Description

Four claims about settlement on this page drifted from the contract, and the reason is structural:
the page carried **17 hand-written Rust blocks against 4 snippets**. Nothing compiled the 17, so
nothing caught them. Fixing the sentences (#304) does not stop the fifth.

The in-memory broker is the project's reference implementation, already tested, and already this
page's source for one snippet. It is now the source for five more, and one hand-written block is
replaced outright:

| Page block | Now |
|---|---|
| lifecycle ladder | `src/memory/mod.rs:ladder` - both transitions, including the `Closed` witness carrying teardown diagnostics the prose describes |
| `Subscribe` | `src/memory/mod.rs:subscribe` |
| `PublishPolicy` + `DefaultPublish` | `src/memory/mod.rs:publish_policy` |
| `FromName` (was a sketch against a broker crate with no compiled home) | `src/memory/mod.rs:from_name` - **replaced**, not supplemented |
| the trait defaults | `src/message.rs:incoming_defaults` |

Snippets per page go from 4 to 9, in all three languages.

## The defaults, and why no new artefact

`MemoryBroker` cannot show what "override nothing" gets you, because it overrides all three - as
does every broker in the workspace. That is the corner the plan allowed a small compiling artefact
for, "only if you cannot pin it another way".

It was already pinned another way. `src/message.rs` has
`incoming_message_defaults_apply_without_override`: a `Stub` implementing `IncomingMessage` with
nothing optional overridden, asserting `partition_key().is_none()`, `!supports_nack_after()`, and
`nack_after(..) == Err(AckError::Unsupported)` - the exact three defaults that drifted, including
the pairing behind error 3. The page now shows that test. `cargo test` runs it, so a default that
moves fails in the core's own suite, and there is no third implementation to keep true.

So: no `tests/minimal_broker.rs`, and no new code at all. This PR adds snippet markers and nothing
else.

## What stays hand-written, and why

Judged against the "noise the page has not introduced" test, not left out for convenience:

- **The trait declarations** (`Broker`, `ConnectedBroker`, `Subscribe`, `Subscriber`,
  `IncomingMessage`, `Publisher`, `PublishPolicy`, `DefaultPublish`, `SubscriptionSource`). These
  are declarations of core traits; no implementation file contains a declaration, so there is
  nothing to cut them from. They stay simplified sketches, as their `inline-rust` markers already
  say, and each now has a compiled implementation directly beneath it.
- **`impl Subscriber for MemorySubscriber`** - about 40 lines of `poll_fn`, seek watermarks,
  stale-copy skipping, waker registration and two `#[cfg(feature = "testing")]` coordinator
  branches, replacing a 6-line sketch. This is the clearest case of a snippet that would need the
  "ignore this part" paragraphs, so the sketch stays.
- **`impl IncomingMessage for MemoryMessage`** - `supports_nack_after` and `nack_after` are indeed
  written out correctly there, but `payload`/`headers` thread an `Option<Delivery>` (the
  ack-consumes-self internal) through a `OnceLock` fallback, `partition_key` delegates to the
  `Partitioned` capability the page introduces 300 lines later, and `nack_after`'s body is a
  `#[cfg(feature = "testing")]` coordinator branch plus a `tokio::spawn`. The defaults test carries
  the teaching this block was wanted for, without the plumbing.
- **`impl Publisher for MemoryPublisher`** - the transaction-buffer branch belongs to
  `TransactionalPublisher`, a capability the page has not reached at that point.
- **The four subscriber/publisher settings extension blocks** (`NatsSubscriber`, the `StartAt`
  variant, `NatsPublish`, the call shape) - the in-memory broker has no settings vocabulary, so
  there is nothing here to cut them from. Still unpinned.
- **The prelude probe and the per-delivery context sketch** - `memory::prelude` and `MemoryContext`
  could plausibly source these; left for a follow-up rather than widened into this PR.

Net: 17 hand-written blocks becomes 16, five of them now shadowed by a compiled implementation
below, and the one corner that caused the outage pinned by an assertion.

## Merge order

Test-merged locally against both open branches touching this page: **#304 and #301 both merge
clean**, in either order. No coordination needed.

## Type of change

## Checklist

      `examples/` where applicable


powersemmi:

PR #306, opened 2026-09-08
Say how to write an in-process transport a test can trust

powersemmi:
# Description

The page told an author to implement `TestableBroker` and stopped. What it never said is how much
of the broker to reproduce - and a service's whole test suite runs against that type, so every
difference between it and the real transport is a green test for behaviour the transport does not
have.

New subsection at the end of `## Test support`, written from the defects that motivated it rather
than from first principles:

- **Run the core's contract suites against the in-process transport, not only against a server.**
  First on the list because `lifecycle` asks the question a stand-in almost never gets asked: does
  a publisher created before `shutdown` fail afterwards? A real client answers "not connected"; one
  whose publish is a channel send has no reason to. Cut from `tests/conformance_self.rs`, which is
  the core running its own suites against its own in-process transport - `run_suite` was already
  marked, `lifecycle` is marked here.
- **Capability parity, ranked honestly.** The `testing` feature is for tests and a release build
  turns it off - stated explicitly, because a reader who does not know that cannot judge the rest.
  Falling short is the direction that costs: the handler cannot be mounted in process at all and
  the behaviour goes untested. Going over is a smaller matter, a compile error in the author's own
  release build.
- **Settle the way the transport settles** - `AckError::Unsupported` where the real `ack` reports
  it, rather than `Ok(())` to keep a suite quiet.
- **Reproduce what is client-side; refuse to fake what is broker-side.** Competing consumers is
  called out as the one to get right, because getting it wrong looks like success: a queue that
  hands every message to every subscriber is not a queue, and a test counting processed messages
  sees the work done and reports nothing.
- **Every gap carries the assertion it makes unsound**, not just "this is missing", plus the suite
  that does cover it.
- **Pin the gap with a test**, since a comment goes stale the first time someone "fixes" the
  transport - a test asserting the handler is *not* reached fails that day.
- **Mount it with the production wiring**, so a service tests the routes file it ships.

## What is prose rather than a snippet

The two code blocks are snippets. The gap-comment example is inline with an `inline-rust`
justification: it is the shape of a comment, and the in-memory broker has no transactional id to be
fenced on, so there is nothing real to cut it from. The behavioural rules are prose because the
in-memory broker's routing carries the coordinator plumbing and seek bookkeeping this section has
not introduced - the same judgement as #305.

## Translations

The first commit calqued the English. The second rewrites both translations against the house
style: the corpus term for the thing being described (`внутрипроцессный транспорт` /
`进程内传输`, already used in `conformance.md`) instead of a coined word, no fan-out-as-picture, no
behaviour that "lives" on a side, no combat idiom for a running service, sentences split where the
English period ran long, and the glossary's locked `at-most-once` verbatim.

## Merge order

Test-merged against all three open branches touching this page - #301, #304 and #305 - **clean in
every case**, in any order.

## Type of change

## Checklist

      `examples/` where applicable


powersemmi:

PR #307, opened 2026-09-08
Sweep the Russian docs for idioms that name nothing checkable

powersemmi:
# Description

A correctness sweep of `docs/ru/`, not a rewrite. The starting point was the two `в бою` lines
flagged during #306 review; the question worth answering was whether that was two lines or a
pattern. It was a pattern: **28 sites across 10 pages**, in four classes.

## What was found, by class

| Class | Sites | Verdict |
|---|---|---|
| Combat idiom for a running service (`в бою`, `боевой брокер`, `боевой код`) | 5 | fixed |
| Transport metaphor for a mechanism (`едет на`, `живёт в`) | 21 | fixed |
| Glossary term carrying a second meaning (`совместимость`) | 1 | fixed |
| Locally invented term where the corpus has a settled one | 0 | already clean |
| Glossary-locked term translated anyway | 0 | already enforced, see below |

**The combat idiom** names nothing a reader can check. English says "in production" / "a production
broker" / "the production code"; the Russian reached for register instead. There is no settled
rendering in the corpus (I grepped for one), so all five sites now use the same plain wording.

**The transport metaphor** is the large one and the one already ruled out once: the style guide
records that `живёт / едет / приезжает` must become the verb of the actual mechanism. It survived
in 21 places. Each is now the verb that says what happens - `передаётся`, `проходит`, `задаётся`,
`действует`, `хранится`, `объявлен`, `подключается` - picked per site, not swapped uniformly, since
"rides on a registration" and "rides in the headers" are different claims.

**`Совместимость`** is the glossary's locked rendering of *capability*. `example-nats.md` used it
for *compatibility* in the SemVer sense ("so new variants are not breaking"), so one word carried
two meanings across the corpus. Now `ломающим изменением`.

## Two classes that came back clean, and why that is worth knowing

- **No locally invented terms.** The `stand-in` coinage I had to fix in #306 does not appear
  anywhere else; `внутрипроцессный транспорт` is used consistently (18 sites). The one `тестовый
  дублёр` is a faithful translation of the English "test double", not a competing coinage.
- **No glossary-locked term was translated.** Worth correcting an assumption in the brief here:
  `check_doc_i18n.py` walks `DOCS.rglob("*.md")`, so it already checks the whole tree on every run,
  not only pages a diff touches. That class cannot rot.

## Parity gate

Russian only - no English page is touched, exactly as asked. The parity rule is symmetric, so it
fails on a single-language edit by design, and the script's own message points at the way out:
the `docs-single-language` label, which `docs.yml` reads to skip that job alone. Applied. This is
the case the label was written for: a word choice that was already right in the other trees.

Test-merged against all five open branches from this thread (#301, #303, #304, #305, #306): clean
against every one.

## Type of change

## Checklist

      `examples/` where applicable


powersemmi:

PR #322, opened 2026-09-09
feat: let a publisher adapter keep the include site's codec

powersemmi:
# Description

A broker that ships a per-publish setting of its own (an ordering key on Pub/Sub, a routing hint
elsewhere) has to wrap the publisher and become the sink for that publish. The wrapper is a bare
`Publisher`, so a builder started on it resolved the crate default codec, and those messages left
in a different format from every other message the same registration sends. Found on
`ruststream-gcp-pubsub`, where `with_ordering_key` silently moved a mount site's `cbor` publishes
onto json; the shape is the same for every broker that adapts a publisher.

`Slot::message_through(sink, value)` starts the entry's own typed publish and sends it through a
sink the caller names. The codec stays the entry's, the adapter's base headers still reach the
wire, and the message still travels the slot's publish path and keeps its test-capture
attribution.

I did not take the shape I first proposed - a trait on the sink that `PublishExt::message` asks -
because it is not expressible on stable. `Slot` is a `Publisher`, so a blanket impl for every
publisher and an override for the entry overlap, and coherence rejects the pair without
specialization. Making the trait required instead would compile-break all twelve broker crates for
no gain on the wire. `Codec` gains no `Default` or `Clone` bound: reconstructing a codec value from
its type would hide the same divergence one level deeper.

Nothing changes format today - this is purely additive and nothing called it before. A broker moves
its adapter onto it one crate at a time, and only the publishes that are already wrong change.

## Type of change

- [ ] Documentation (typos, code examples, or any documentation updates)
- [ ] Breaking change (removing or renaming public API, changing a signature, tightening bounds, or

## Checklist

      `examples/` where applicable

### DIFF
--- a/src/runtime/handle/outs.rs
+++ b/src/runtime/handle/outs.rs
@@ -88,7 +88,7 @@ use crate::runtime::inject::FromStartup;
 use crate::runtime::inject::{InjectCall, InjectDef};
 use crate::runtime::metadata::OutgoingMessageMetadata;
 use crate::runtime::publish::{
-    Admits, HeadersUnset, MessageBody, OutPipeline, PublishBuilder, PublishIdentity,
+    Admits, HeadersUnset, MessageBody, OutPipeline, PublishBuilder, PublishIdentity, PublishSink,
     TransactionScope, TypedTransaction, message_of,
 };
 use crate::runtime::router::IncludeDef;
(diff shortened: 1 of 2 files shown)


powersemmi:
Closing in favour of the real fix. The divergence this branch worked around - a publish through a broker's publisher adapter losing the codec the mount site named - is a symptom: a broker has nowhere to put per-message settings except a new publisher, because `base_headers` is per publisher rather than per message. Wrapping the publisher is what loses the slot entry. `Slot::message_through` added a second way to start a publish instead of removing the reason to wrap, and so did the alternative that returned a substituted slot. Typed per-message options on the publisher, defaulted by the publish policy, remove the adapters and the divergence together; filed as its own issue.

PR #325, opened 2026-09-09
fix(macros): let a subscription source name the type it produces

powersemmi:
# Description

`#[subscriber(..)]` read the source type off the constructor at the head of the expression and took
every later step to return it. A broker whose subscription descriptor is built on typestates could
therefore not be written in the attribute at all: the chain ends on a type the attribute has already
decided against, and the user sees a mismatch reported against generated code. Typestates are the
technique this project asks for when two delivery models share no settings, so this made the
preferred technique unavailable to broker authors on the path their users actually use. Two broker
crates carry a shape they did not choose because of it.

`<expression> as <Type>` now names what the source produces. It parses as a cast, which is free
tokens here since a cast between two subscription descriptors is not something anyone writes. The
same form takes in a source built by a free function, which the macro rejected outright before, and
a `start_at(..)` position whose tokens do not carry its type.

Inferring the type instead is not open to a macro, and the shape proposed for it,
`<expr as IntoSource>::Source`, is not Rust: the left of a qualified path must be a type, and
naming the type of an expression has no spelling on stable. The type is load-bearing in
`Declared::Settings`, an associated type the expansion has to write, and it travels from there into
`RouterMount::Out`, so the mount chain's own type depends on it. Erasing it would mean an opaque
mount surface, which is a different change and not a bug fix. So the guess stays and is now checked
rather than assumed: a wrong one is a compile error that names the type the expression actually
produces and the form that fixes it.

Every form that compiled before still compiles: a string literal, a bare constructor, a struct
literal, and a chain of same-type steps are unchanged, and the two open forms never guessed at all.
`tests/macro_subscriber.rs` gains a descriptor whose chain moves the value to a second type,
mounted through the attribute and driven end to end; the memory broker has no second state to model,
carrying a name and nothing else, so the case lives in the test surface rather than on the broker.

Two UI snapshots grew by one clause, the one that names the way out. Three others in the tree
(`lane_typo_deserialize`, `registration_layer_blanket_only`, `reply_codec_named_twice`) are stale
against a newer rustc for reasons unrelated to this change; I left them as they are rather than
refresh them here.

## Type of change

## Checklist

- [x] New and existing tests pass locally (`just test`, with `RUN_UI_TESTS=1`)


powersemmi:
Closing: the case this opened up is settled the other way. A broker with several subscription forms gives each form its own descriptor carrying that form's settings, so a required setting is a constructor argument and every chain step returns Self. A chain that changes type mid-way is then not a shape this project writes, and naming the type by hand has nothing to enable. The two side cases it also allowed - a source returned by a free function, and a start_at position whose type the tokens do not carry - are conveniences nobody asked for, and both fail today with a clear error rather than silently. The branch stays if the question reopens.

PR #337, opened 2026-09-12
feat: wire the deferred-retry publisher from a policy, paired on the connected broker

### LINKED ISSUE #335: Wire the deferred-retry publisher from a policy, paired on the connected broker
### Problem

`BrokerScope::retry_via` takes a live `Publisher`. A broker that follows the lazy-startup contract has no live publisher before `connect`, and the runtime calls `connect` only after the synchronous builder has run. So on such a broker the deferred-retry fallback cannot be wired at all: the subscription now reports where a deferred copy is published (#321), and the service has nothing to publish it with. The rc.3 adoption wave hit this on RabbitMQ and on the Redis list and pub/sub forms; every broker that ships no cell-backed early publisher is in the same position.

### Proposed solution

`retry_via` takes the broker's publish policy instead of a live publisher, and the runtime pairs it on the connected form at startup, next to where it pairs the reply policies:

```rust
b.retry_via(Publish::default());
```

The scope stores the policy until `connect` has run, pairs it once, and holds the live publisher erased exactly as today (`ErasedPublisher`, the deferred copy publishes through it with no options). The pairing error surfaces as a startup error naming the scope, like a reply policy that fails to pair. The variant taking a live publisher goes: a policy is the one form every broker has, and two entry points for one wiring would leave the broken one in place.

### Alternatives considered

**A cell-backed early publisher in every broker.** The contract allows it but does not make it the canon; it would mean twelve copies of the same `Arc<OnceCell<_>>` for one runtime need. Rejected.

**Keeping both `retry_via(publisher)` and a policy form.** Two ways to wire one thing, and the publisher form stays unusable on every lazy broker. Rejected.

### Additional context

Breaking: the signature of `retry_via` changes. Pre-release, lands in the 0.7 line. Broker crates that document "retry_via cannot be wired on this broker" (lapin, fred) drop that sentence once they adopt the core that carries this.

powersemmi:
# Description

`BrokerScope::retry_via` takes the broker's publish policy (`b.retry_via(Publish::default())`) and the runtime pairs it on the connected broker at startup, before the scope's first subscription opens. Every subscription of the scope shares the one live publisher, held erased as before, and a policy that fails to pair aborts startup the way a reply policy does. Nothing changes for a broker with native delayed redelivery.

Beyond the issue: the pairing runs lazily on the first subscription the scope opens rather than in a new startup phase, so `run` keeps its sequence; the unit tests prove one pairing per scope and a clear error for a slot startup never filled. The guides in three languages now say the argument is the policy.

Deferred: the broker crates that documented "retry_via cannot be wired on this broker" (lapin, fred) drop that sentence once they adopt the core carrying this.

Fixes #335

## Type of change

## Checklist

      `examples/` where applicable


powersemmi:
Superseded: the deferred-retry publisher becomes a position on the mount chain (`.out(Retry, policy)`, with the `out_retry` sugar) instead of a scope method; see the follow-up issue.

PR #340, opened 2026-09-12
docs: fix the NATS example's code and the translation banner

powersemmi:
# Description

The NATS example page taught code that does not compile: the descriptor called `connected.subscribe(self)` where the connected form's `subscribe` takes a name and `subscribe_with` takes the descriptor, the prelude block exported a `NatsSource` the page never defines and the seek capabilities the page says NATS does not implement, and one import was unused. All three languages carry the same blocks and are fixed alike. Two smaller things ride along: the Russian broker table names Kafka's offset commits with the Kafka crate's own term, and the Russian translation banner says what the Chinese one already said, that the translation was produced by a model.

The `docs-single-language` label covers the Russian-only table wording; the example fix reached every language.

## Type of change

## Checklist


powersemmi:
Folded into #339 at the maintainer's request: one pull request for the core, as for every broker repo. The branch stays until #339 merges.

PR #356, opened 2026-09-13
feat: pass the resolved destination to the publish-side bindings hooks

### LINKED ISSUE #352: Pass the resolved destination to the publish-side bindings hooks
### Problem

The publish-side bindings hooks of #345, `PublishPolicy::channel_bindings`, `operation_bindings` and `message_bindings`, take `&self` alone, so a policy cannot name the channel it publishes to. The rc.5 adoption of the SQS/SNS crate stopped there: the `sns` channel binding's only required field is the topic `name`, the `sqs` binding on the publish side needs the queue's `name`, and the policy knows neither, because the destination is resolved at the mount site (a reply type's `#[outgoing(name)]`, a `publish("dest")` clause, a slot's dictionary, a `dead_letter("dlq")` declaration). A binding without its required field is a document tools refuse, so the crate emits none.

### Proposed solution

The three publish-side hooks receive the resolved destination:

```rust
fn channel_bindings(&self, channel: &str) -> Bindings;
fn operation_bindings(&self, channel: &str) -> Bindings;
fn message_bindings(&self, channel: &str) -> Bindings;
```

The runtime passes the name it reports as the channel's `address` (the mount site's resolved destination), and `null` addresses (a naming transform) pass the mount site's fallback name. `SubscriptionSource`'s three hooks keep `&self`: a descriptor knows its own name. `reply_address_location` keeps `&self`. Every impl in the core and the memory broker gains the parameter; the broker-author guide's bindings section shows the SNS topic name filled from it.

### Alternatives considered

- A separate `named_channel_bindings` hook beside the old three: rejected, two hooks for one binding.
- Reading the destination from the policy's own field: rejected, a policy declares a broker's settings and never a destination (#343's rule); the name lives at the mount site.

### Additional context

Breaking for broker authors within the pre-release: the hook signatures change; the crates that fill publish-side bindings update in their next adoption. Follows #345. 0.7 line.

powersemmi:
# Description

The three publish-side bindings hooks now take the destination the mount site resolved, so an SNS
topic name and an SQS queue name reach the document instead of a binding the crate has to drop.

Fixes #352

Details worth knowing beyond the issue:

- A slot's dictionary names several destinations and one policy answers for each of them, so a
  slot is now described once per destination rather than once per slot.
- The retry position holds its bound policy erased instead of a closure over it: the dead-letter
  channel is named by the registration's own `dead_letter(..)` declaration, which is read after the
  position is bound, and the document asks the policy only then.
- A naming transform leaves the channel with no address, and the hooks are handed the mount site's
  fallback name - the name a delivery the transform left alone is answered on.
- The broker-author guide's bindings example is now an SNS policy filling the topic name from the
  parameter, in all three languages.

## Type of change

## Checklist

- [x] New and existing tests pass locally (`just test`, and the UI tests with `RUN_UI_TESTS=1`)
      `examples/` where applicable


powersemmi:
Folded into #355, which carries this branch merged unchanged on top of main together with #358 and the docs restructure.

PR #358, opened 2026-09-13
feat: carry a bare-name registration's retry declaration to the broker

### LINKED ISSUE #351: Let a bare-name subscription carry the retry declaration to a broker that moves the message itself
### Problem

A registration mounted by a bare name, `#[subscriber("orders-workers")]` or `b.include(h)` over a `Name` source, can declare `max_attempts(n)` and `dead_letter("dlq")` like any other, but on a broker whose subscriptions are `BrokerMoves` the declaration reaches nobody: the core's `Name` source takes the broker's copy path from `Subscribe::Copies`, its `declare_retry` is the default that records nothing, and `Subscribe` has no hook a broker could apply the declaration through. The rc.5 adoption of the Pub/Sub crate found it: the cap declared over a bare subscription name is silently not applied, and a `dead_letter` without `max_attempts` over such a name makes the first redelivery a `nack(false)` that a subscription without a dead-letter policy acknowledges, losing the message. The same shape exists on SQS and Pulsar. Separately, the broker-author guide names Pub/Sub both as the example of `BrokerMoves` and, in the section on where a copy is published, as a descriptor that asks the API for its topic, which the rc.5 crate no longer does.

### Proposed solution

`Subscribe` gains a defaulted method the `Name` source calls where a descriptor calls `declare_retry`:

```rust
fn declare_retry(&self, name: &str, declaration: &RetryDeclaration) -> Result<(), Self::Error>;
```

The default accepts an empty declaration and refuses a non-empty one on a broker whose `Copies` is `BrokerMoves`, with an error naming the broker, the name and the fix ("declare the cap and the destination on this broker's own descriptor, which maps them onto the subscription"); on `AddressedCopies` and `NamedCopies` brokers the default accepts, because the runtime applies the declaration there. A broker with a native mechanism implements it for the bare name the way its descriptor does (Pub/Sub applies the dead-letter policy to the named subscription, SQS the redrive policy to the named queue, Pulsar the `DeadLetterPolicy` to the consumer), and refuses half a declaration the same way. The refusal is a startup error, not a compile error: the name form is dynamic by construction, which is the external force the "why" comment names. The broker-author guide's section on where a copy is published drops the Pub/Sub example and uses a descriptor that stays `AddressedCopies`.

### Alternatives considered

- Leaving the bare name without a hook and documenting that a declaration needs a descriptor: rejected, the declaration compiles and silently does nothing, and one of its halves loses messages.
- Refusing every declaration over a bare name on every broker: rejected, on an open path the runtime applies it correctly today.
- A compile-time refusal: rejected, the `Name` source's copy path comes from the broker and the name is a string; the mount site cannot see whether this broker maps declarations for bare names.

### Additional context

Additive for broker authors (a defaulted method) and for service authors; the three crates with native mechanisms implement it in their next adoption. Follows #343. 0.7 line.

### LINKED ISSUE #354: Cut the site down to its entry pages and move the guides into the module overviews on docs.rs
### Problem

The site carries 87 pages in three languages. Seventeen of them are guides that walk the runtime surface page by page, each kept three times, and the translation parity gate turns every API change into a three-file edit. A reader of a Rust crate goes to docs.rs and to the first-project page, and the module overviews they land on there are nearly empty (`runtime` opens with one line): the place that is read says the least, and the place that is not read costs the most. The same narrative is kept a fourth time in the committed Cursor rules.

### Decision

- The site keeps its entry pages, in all three languages and under the parity gate as today: the home page, installation, quick start, tutorial, the broker index and the memory broker page, the four broker-authors pages, benchmarks, and the API-reference stub.
- The seventeen guides leave the site. Each becomes the overview of the module that owns its topic on docs.rs: `runtime` (subscribers, publishing, routing, context and state, typed headers, lifecycle, middleware, failure policy, running beside another server), `codec`, `testing`, `asyncapi`, `otel`, `metrics`, `logging`, `runtime::cli`, plus a reading order at the crate root. An overview is a compression of the guide, one to a hundred lines per topic with one compiling example, not a transfer of its text; the item docs already carry the detail. Every example is a doctest gated on the features it needs, so it compiles on both feature edges.
- The `runtime` overview is a markdown file next to the module, pulled in with `include_str!`, so `mod.rs` stays within the file-size limit and the text stays prose for the tooling that indexes the repository.
- Old guide URLs redirect to the docs.rs module pages through the redirects plugin already in use. The 31 links from the remaining pages, the README link and the broker sites' links (27 of them to the testing guide) are retargeted to docs.rs; the broker sites retarget in their next adoption wave.
- `context7.json` gains a `rules` list with the framework's decisions, the layer the "Ask AI" widget prepends to every answer. The Cursor rules stay as the agent copy: two copies of the narrative instead of four.

Two pull requests: the overviews first, then the cut with the redirects and the retargeted links, so the redirects have targets.

### Alternatives considered

- Deleting the guides and relying on the item docs alone: rejected. What an item cannot say is which of the legal paths is the recommended one (macro or manual, sugar or marker form, name or descriptor), and that is the first thing a reader asks.
- Moving the guides into rustdoc as they are: rejected. Four thousand lines of prose belong on a site, and a module page that long is not read.
- Dropping the site entirely, README plus docs.rs: rejected. The translations of the entry pages stay, and docs.rs is English only.
- Dropping the translations: rejected for the entry pages, which are the pages usage analytics of comparable documentation show being read.

### Additional context

Two things are lost, on purpose. The overviews are published with a crate release, not with every merge to `main`. They have no translation, so a reader of the Russian or Chinese site gets the entry pages in their language and the English rustdoc after.

powersemmi:
# Description

A bare-name registration's retry declaration now reaches the broker, and the broker answers for it.

Fixes #351

The method the issue sketches takes an error the core can build: a defaulted body cannot construct
a broker's own `Error`, so the signature is
`declare_retry(&self, name, declaration) -> Result<(), DeclareRetryError>`, with `DeclareRetryError`
shaped like `AckError` - an `Unsupported` variant the default answers with, and a `Broker` variant
that carries an implementation's own refusal (a broker that refuses half a declaration). The
refusal reads as one startup line naming the subscription, the broker type and where the
declaration belongs.

The `Name` source reaches the broker through a second defaulted method on `SubscriptionSource`,
`declare_retry_on(&self, connected, declaration)`, which the runtime calls where a descriptor is
handed `declare_retry`. A descriptor of your own never implements it; the decorators (`Buffered`,
`StartAt`) forward it. That method is what makes the broker's answer reachable from a generic
resolve path at all, and it is additive for broker authors like the first.

The copy path decides the default's answer from the type: the sealed trait behind `CopyPath` gained
the question, so `AddressedCopies` and `NamedCopies` accept and `BrokerMoves` refuses with no
runtime inspection of a kind.

Deferred: the `Subscribe` section of the broker-author guide still teaches `NamedCopies` with a
Pub/Sub subscription as its example, which the rc.5 crate no longer answers; the issue named only
the copy-destination section, so that one is left for the maintainer to rule on.

## Type of change

## Checklist

      `examples/` where applicable


powersemmi:
Folded into #355, which carries this branch merged unchanged on top of main together with #356 and the docs restructure; the conflict with #359 in tests/retry_declaration.rs is resolved there, both test sets kept.

